Issue #245, opened 2013-04-23
Support comma first style of variable declaration


mrchief:
Allow code to be formatted like this:

```
var a = 1
  , b = "somethign else"
  , isAwesome = true;
```

And apparently, we have a somewhat ready [patch](http://stackoverflow.com/questions/10872184/comma-first-js-formatter) available. Would be great if this can get included in here!

olsonpm:
+1 for comma first support, and I agree with lewisdiamond's sentiment that ensuring tabs line up properly should not be the end goal as there are pragmatic reasons for the comma-first style.  I would love support for json as well.  Currently I have a hack in the global js-beautify code to do a lot of punctuation-first alternatives (for ternary operators was a must).

bitwiseman:
To everyone who +1'd this: please take a look at branch `feature/comma-first` - https://github.com/beautify-web/js-beautify/commit/e5bfcf34b6089752e0a474e86ff533b46ac1677a

Tell me if this is a sufficient first step towards what you want that I should add it to the next release.

bitwiseman:
@olsonpm, @lewisdiamond, @lukemartin, @mrchief - could you guys take a look at branch https://github.com/beautify-web/js-beautify/tree/feature/comma-first .

olsonpm:
I will spend 30 minutes on it tonight and respond with thoughts.  Thanks for following up.

olsonpm:
Got around to it this morning.  It all looks good to me due to your tests.  I have a personal branch with a lot of operator first functionality - I put my comma-first tests through your branch and they all passed no problem which is great news.

 - When I created my personal branch, I decided the comma-first operator would be passive, meaning the  following:

```
var a,
    b;
```

would not be changed.  You clearly stated this in the commit message, I just feel it's worth discussing with others what behavior they'd expect if they chose to use opt.comma_first.

Another thing, <Output>.previous_line and flags.previous_text are the same thing I believe, which was confusing to me at first but made sense in how you used it.

Other than that, the main difference between our implementations is that you opted to modify the previous line.  I was trying not to "go back and edit things" because I thought it was more confusing to develop.  Now your way is more concise than mine - it is easily read and you commented everything.  The existing code also edits previous tokens so your code is perfectly in line - honestly I'm just bringing it up because I'd like to know what your thoughts are.  In short, my opinion is that this program would be much easier to debug if tokens were only in charge of the white space ahead of them based on next tokens.

Thanks for addressing this!

[...]

PR #369, opened 2013-12-17
Support comma-first for vars/object declarations

tgirardi:
Support "comma first" style when declaring variables or properties
inside an object, as discussed on issue #245.

The proposed solution is based on changing the order of the newline
(before / after the comma) depending on the value for the new option
"comma_first". The proposed command line argument for this option
is "-F" or "--comma_first").

It is also necessary to modify the indentation string used for some
particular lines, in order to include the comma as part of the
indentation string and achieve the desired result regarding
indentation (in some cases it is easier to think of the comma as part
of the indentation string, in replace of a whitespace).

TODO 1: tests for this feature

TODO 2: implement the feature in the python version (I prefer to get
feedback for this commit first and, once the feature's JS code is
considered correct, then proceed with the py code).

TODO 3: in the future, other variations for the indentation pattern,
for object properties, could be allowed. The currently allowed one
looks like this:

```
{
  prop1: 1
, prop2: 2
, prop3: 3
}
```

Another version could be:

```
{
    prop1: 1
  , prop2: 2
  , prop3: 3
}
```

In "other words", the first property is double indented. However, it
must be noted that jslint errors related with the indentation could
arise (tested with JSHint).


bitwiseman:
To everyone interested in this: please take a look at branch `feature/comma-first` - https://github.com/beautify-web/js-beautify/commit/e5bfcf34b6089752e0a474e86ff533b46ac1677a

Tell me if this is a sufficient first step towards what you want that I should add it to the next release.

bitwiseman:
@tgirardi, @chriscorwin @andrew-luhring - Could you look at the branch above and comment?

PR #574, opened 2014-11-10
 CSS: support add newline between rules

numbbbbb:
It's suggested by [Google](https://google-styleguide.googlecode.com/svn/trunk/htmlcssguide.xml#CSS_Formatting_Rules) that newline should be added between rules, so I think we should support
that and try to make it.

I added this behavior to both JS and Python with tests. Besides, I added
the new option to README and cli.js.

If I need to modify other files, please let me know.

Thanks!


joseluisq:
Great news ! 
So I can't wait for the next 1.5.5 version :+1: 
Congrats :smile: !

joseluisq:
I hope the Sass support for the next version..

PR #611, opened 2015-01-30
Allow bool op first

olsonpm:
allowing boolean opoerator-first notation via option 'allow_bool_op_first'.  Only changing node version for now as python can wait until changes are final.

### DIFF
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -206,8 +206,8 @@
             'TK_EOF': handle_eof
         };
         
-		// sub-types
-		var boolOps = ['&&', '||', '?', ':'];
+        // sub-types
+        var boolOps = ['&&', '||', '?', ':'];
 
         function create_flags(flags_base, mode) {
             var next_indent_level = 0;
@@ -868,12 +868,12 @@
             if (last_type === 'TK_COMMA' || last_type === 'TK_START_EXPR' || last_type === 'TK_EQUALS' || last_type === 'TK_OPERATOR') {
                 // taking care of ambiguous case allow_bool_op_first case where a variable is on its own line after an operator
                 var nextTok = get_token(1);
-				var ambiguousBoolOpCase = opt.allow_bool_op_first &&
-					last_type === 'TK_OPERATOR' &&
-					in_array(flags.last_text, boolOps) &&
-					current_token.wanted_newline &&
-					nextTok.wanted_newline;
-				
+                var ambiguousBoolOpCase = opt.allow_bool_op_first &&
+                    last_type === 'TK_OPERATOR' &&
+                    in_array(flags.last_text, boolOps) &&
+                    current_token.wanted_newline &&
+                    nextTok.wanted_newline;
+                
                 if (!start_of_object_property() && !ambiguousBoolOpCase) {
                     allow_wrap_or_preserved_newline();
                 }
@@ -1095,7 +1095,7 @@
             // Support ternary newline preservation
             // a = (b === c)
             //     ? c
-            //	   : d;
+            //     : d;
             //
             // and also boolean operators
             // a = (b === c)
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -205,6 +205,9 @@
             'TK_UNKNOWN': handle_unknown,
             'TK_EOF': handle_eof
         };
+        
+		// sub-types
+		var boolOps = ['&&', '||', '?', ':'];
 
         function create_flags(flags_base, mode) {
             var next_indent_level = 0;
@@ -270,6 +273,7 @@
         opt.wrap_line_length = (options.wrap_line_length === undefined) ? 0 : parseInt(options.wrap_line_length, 10);
         opt.e4x = (options.e4x === undefined) ? false : options.e4x;
         opt.end_with_newline = (options.end_with_newline === undefined) ? false : options.end_with_newline;
+        opt.allow_bool_op_first = (typeof options.allow_bool_op_first === 'undefined') ? false : options.allow_bool_op_first;
 
 
         // force opt.space_after_anon_function to true if opt.jslint_happy
@@ -862,7 +866,15 @@
             }
 
             if (last_type === 'TK_COMMA' || last_type === 'TK_START_EXPR' || last_type === 'TK_EQUALS' || last_type === 'TK_OPERATOR') {
-                if (!start_of_object_property()) {
+                // taking care of ambiguous case allow_bool_op_first case where a variable is on its own line after an operator
+                var nextTok = get_token(1);
+				var ambiguousBoolOpCase = opt.allow_bool_op_first &&
+					last_type === 'TK_OPERATOR' &&
+					in_array(flags.last_text, boolOps) &&
+					current_token.wanted_newline &&
+					nextTok.wanted_newline;
+				
+                if (!start_of_object_property() && !ambiguousBoolOpCase) {
                     allow_wrap_or_preserved_newline();
                 }
             }
@@ -1076,6 +1088,26 @@
             // if there is a newline between -- or ++ and anything else we should preserve it.
             if (current_token.wanted_newline && (current_token.text === '--' || current_token.text === '++')) {
                 print_newline(false, true);
+            }
+            
+            // opt.allow_bool_op_first
+            //
+            // Support ternary newline preservation
+            // a = (b === c)
+            //     ? c
+            //	   : d;
+            //
+            // and also boolean operators
+            // a = (b === c)
+            //     && d
+            //     || e
+
+            if (opt.allow_bool_op_first && in_array(current_token.text, boolOps)) {
+                if (last_type === 'TK_END_EXPR' || last_type === 'TK_WORD' || last_type === 'TK_STRING') {
+                    if (!start_of_object_property()) {
+                        allow_wrap_or_preserved_newline();
+                    }
+                }
             }
 
             // Allow line wrapping between operators
--- a/js/test/beautify-javascript-tests.js
+++ b/js/test/beautify-javascript-tests.js
@@ -13,7 +13,8 @@ function run_javascript_tests(test_obj, Urlencoded, js_beautify, html_beautify, 
         space_before_conditional: true,
         break_chained_methods: false,
         selector_separator: '\n',
-        end_with_newline: false
+        end_with_newline: false,
+        allow_bool_op_first: false
     };
 
     function test_js_beautifier(input)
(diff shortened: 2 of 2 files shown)


bitwiseman:
Sorry for the long wait, if you're still willing to do this, I'll be more responsive. 

If we're going to do this we should simply force, but make the option for all operators.  Call it 
`preserve_operator_newline`, for example.  And allow it to be set to `after`, `before`, or `both`. 

This would be a simpler change - no new class of operators... 

It looks like we already allow line wrapping between operators:
https://github.com/beautify-web/js-beautify/pull/611/files#diff-655cdbdfd52b6516fe4d5d2e53a7af50R1114

...


Regarding  tests: 
What you have is good but we use a central data file to generate tests.  it will do what you're trying to do at https://github.com/beautify-web/js-beautify/pull/611/files#diff-62923ff6f2353cc359176aa9e7d376e1R1709 , but generating the same method calls twice using templating.  Look at https://github.com/beautify-web/js-beautify/commit/f8cf725b096439179ca1cb08e31f0ee7ca4eb577 for an example.  You can use the same inputs and have the outputs templated.  

Run `make testj` to regenerate and run tests.

olsonpm:
Thanks much.  I'm hoping to have time this weekend to make your suggested improvements.  If not then for sure some time next week.

olsonpm:
> If we're going to do this we should simply force, but make the option for all operators. Call it 
preserve_operator_newline, for example. And allow it to be set to after, before, or both.

Re-reading your comment, please elaborate on the above.

My understanding of "we should simply force" is that operators should be forced to be before newlines or after newlines.  Additionally, users should be able to keep operators 'as-is', without being forced into either.  If my understanding is correct, we should then find some way to word it appropriately, because forcing might add or subtract a newline as opposed to preserve it.

Anyway, I'll start coding toward an option 'operator_first' with possible values of 'force', 'allow', 'false' -> false being the default.  I'll await your thoughts before creating another PR.

olsonpm:
Also, if you have a data file that then generates the actual tests, why version the tests themselves?

olsonpm:
Alright - final comment for now:

I'm unsure how to go about writing tests if we're to write 'operator_first' functionality.  The number of operators and cases that would ideally be tested are unreasonably large.  At least from what I can tell, operators are defined as all punctuation that isn't a comma or an equality symbol:

```javascript
var punct = ('+ - * / % & ++ -- = += -= *= /= %= == === != !== > < >= <= >> << >>> >>>= >>= <<= && &= | || ! ~ , : ? ^ ^= |= :: =>'
                +' <%= <% %> <?= <? ?>')
```

along with [line 1990](https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L1990-L1996)

Can you give me some idea as to what you're expecting?

bitwiseman:
> Also, if you have a data file that then generates the actual tests, why version the tests themselves?

The data file generates the actual tests but it too is code, and migration/implementation is just now reaching the point where it is stable. I totally plan to remove the generated files from the repository - it causes (very justified) confusion with every single person that sees it.

bitwiseman:
> Can you give me some idea as to what you're expecting?

I agree naming this is hard.  I'm going to use `operator_positioning` for this example. 
```js
var d = 1;
if (a === b 
    && c) {
    d = ((c * everything / something_else)  
            - a) %
        b;
    e 
        += d;

} else if (!(complex && simple) ||
        (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many ||
        anOcean ||
        aRiver);
}

// Output: actual current 
var d = 1;
if (a === b && c) {
    d = ((c * everything / something_else) - a) %
        b;
    e
        += d;

} else if (!(complex && simple) ||
    (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many ||
        anOcean ||
        aRiver);
}

// Output: operator_positioning = preserve_newline
var d = 1;
if (a === b 
    && c) {
    d = ((c * everything / something_else)  
            - a) %
        b;
    e 
        += d;

} else if (!(complex && simple) ||
        (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many ||
        anOcean ||
        aRiver);
}

// Output: operator_positioning = after_newline
var d = 1;
if (a === b
    && c) {
    d = ((c * everything / something_else)
            - a)
        % b;
    e 
        += d;

} else if (!(complex && simple)
        || (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many 
        || anOcean
        || aRiver);
}

// Output: operator_positioning = before_newline
var d = 1;
if (a === b &&
    c) {
    d = ((c * everything / something_else) -
            a) %
        b;
    e +=
        d;

} else if (!(complex && simple) ||
        (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many ||
        anOcean ||
        aRiver);
}

// Output: operator_positioning = remove_newline
var d = 1;
if (a === b && c) {
    d = ((c * everything / something_else) - a) % b;
    e += d;

} else if (!(complex && simple) || (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many || anOcean || aRiver);
}
```

Very rough, I'm open to discussion.  For example, maybe the first operator in a statement (the `+=` in `e += d`) should never have a newline before it... 

```js
a 
    += 
    b = 
    c += d;

// So, even with 'operator_positioning = preserve_newline'
// shouldn't the the above would at least reformat to below
a  += 
    b = 
    c += d;
```

olsonpm:
operator_positioning makes sense.

> For example, maybe the first operator in a statement (the += in e += d) should never have a newline before it

I also wasn't sure how to properly define operators that would and wouldn't be placed after a newline.  I'll give it some more thought and figure out a suggestion by tonight/tomorrow.

bitwiseman:
Okay, one change to what I said: `operator_position` not `operator_positioning`.  

I've also just pushed a change the removes the old tokens used to recognize template language tokens (we do it a different and better way now).

olsonpm:
Alright - things have been coming up at home but I think I've wrapped my mind around this problem.

Considering the following 'master list'
```javascript
'+ - * / % & ++ -- = += -= *= /= %= == === != !== > < >= <= >> << >>> >>>= >>= <<= && &= | || ! ~ , : ? ^ ^= |= :: =>'
```

Categorically speaking, while disregarding = and , for TK_EQUALS and TK_COMMA:
```
arithmetic operators
+ - * / %

in/decrement
++ --

bitwise operators
& | ^ ~

assignment operators
= += -= *= /= %= >>>= >>= <<= &= ^= |=
ambiguous case ':'

comparison operators
== === != !== > <

boolean operators
&& ||

conditional operators
?
ambiguous case ':'

arrow function
=>
```

The above is important for considerations in practical usage in javascript.

Considering newline placement, we can ignore assignment operators as I can't see where it would ever make sense for them to be on their own line.  Unary operators such as ++ and ~ should also never be placed by js-beautify.  I have not worked with arrow functions in javascript, but if my experience with lambdas in c# is worth anything, I don't see a use-case where its placement should be modified.  And conditional operators (the only ternary operator in a lot of languages from what I read) should definitely be placed on their own line. which brings us to the following:
```
binary operators
+ - * / % & | ^ && || == === != !== > <

conditional operators
? :

ambiguous :
```

Now I don't feel comparison operators should be on their own line, but this is from personal experience in navigating code.  I've just never seen it.  Strictly speaking, it is a binary operator that outputs a boolean, which is identical to && and ||.  Conceptually however they don't read in a fluent manner.  I'm open to discussion on this.

Last but not least, I think we'll need to treat : as a special case, considering its use both inside object notation and conditional operators.


tldr, my suggestion
```
modify placement of
+ - * / % & | ^ && || ? :
```

bitwiseman:
I'd say put the comparison operators in the list too.  If someone put a newline on either side of those, we should respect that even if it seems idiotic.  This project has truly taught me that whatever I think is reasonable there a contingent that thinks the opposite.  

Also you dropped the bitwise shift operators from for all the lists before starting.

olsonpm:
Ah - don't know how I missed those.

I definitely trust your judgement on the comparison operators.  Since the bitwise shifts are binary operators, I assume you want them positioned as well?

olsonpm:
Hey - finished a first-draft:

https://github.com/olsonpm/js-beautify/tree/add-operator-position

Let me know what you think.  Two comments in particular I'm interested in:

1) 'after_newline' as defined in your tests might not be what we want.  Specifically:

```javascript
var a = b
+
c;
```
will resolve to
```javascript
var a = b
   + c;
```
as opposed to what I think should be:
```javascript
var a = b
    +
    c;
```

2) In order to prevent newlines from being added after operators for both the after_newline and remove_newline options, I modified the following lines:

https://github.com/olsonpm/js-beautify/blob/add-operator-position/js/lib/beautify.js#L1735-L1741

I was confused on how to go about this, and I'm not confident this is how you would have done it.


The only other question I have is about the raw testing functionality.  I couldn't understand what that was for - so if you could give me an overview that'd be awesome.


Thanks,
Phil

bitwiseman:
> 1) 'after_newline' as defined in your tests might not be what we want.

No, for `after_newline` I do want: 
```javascript
var a = b
+
c;
```
to resolve to
```javascript
var a = b
   + c;
```
If someone wants to keep that original structure, they should use `preserve_newline`.  

Also, for `remove_newline`, the input should be converted to:
```javascript
var a = b + c;
```

> 2) In order to prevent newlines from being added after operators for both the after_newline and remove_newline options, I modified the following lines:

No, definitely don't do that.  :smile:   The tokenizer, which should not filter out newlines it should just read the inputs.  About the only thing it should drop is: convert all newline characters to `\n`, and convert whitespace on otherwise empty lines to `\n`.  Other than that, you must be able to accurately reproduce the input from the token stream. 

So, the change you want to make will need to be in the `handle_operator` section, but will also need to add a property similiar to `output.space_before_token` that indicates that user requested newlines before the next token should be ignored... `output.ignore_newline_before_token` maybe.   (That could be used in some other places to prevent us having to trim lines out of the output, too.)

bitwiseman:
The `raw` testing is meant to guarantee that the tokenized input can round-trip to the same output.  It helps ensure the `preserve` directive will work in all cases without me having to write tests specifically for the all the inputs the preserve directive might encounter.

olsonpm:
> If someone wants to keep that original structure, they should use preserve_newline. 

Am I doing something wrong then?  The default behavior (sans preserve_newline) as I'm testing will resolve
```javascript
var a = b
+
c;
```
to
```javascript
var a = b +
    c;
```
And the operator_position as currently written behaves the way you're expecting.

Concerning #2, I will rewrite the logic as you suggest and follow up with a separate, more appropriately titled PR.  Thanks for the feedback.

olsonpm:
I'm stumped and can't move forward until you explain the following case:

why does
```javascript
var a = (b /
c);

// resolve to
var a = (b /
    c);

// and not
var a = (b /
        c);
```
the indentation logic that goes with 'preserve_newlines = true' is breaking my operator placement indentation.  Due to it, the following test:
```javascript
var a = (b
/ c)
% d;
```
resolves to varying indentation lengths depending on which operator_position option is chosen.

Now I can rewrite the tests to account for this, but I have to know why this indentation logic was chosen before doing so.  It just seems incorrect.

olsonpm:
On second thought, I guess most people would expect something like
```javascript
(function() {
    var indent = 1;
});
```
to only display one level of indentation.  I'll rewrite the tests and have the pr by tomorrow.

bitwiseman:
Sounds good. Sorry for the delayed response.  
Thanks for persevering.

olsonpm:
continue thread on #735

PR #613, opened 2015-02-01
rewrite file only when file data changed

### LINKED ISSUE #597: only print the file names of changed files
for now, when I run `js-beautify` from cli, it will print all the file names passed to `js-beautify`, but some are not changed.

So, it's better to print the names of `changed files only`.

coderhaoxin:
also fix #597

### DIFF
--- a/js/lib/cli.js
+++ b/js/lib/cli.js
@@ -278,39 +278,41 @@ function processInputSync(filepath) {
         });
 
         input.on('end', function() {
-            makePretty(data, config, outfile, writePretty);
+            makePretty(data, config, filepath, outfile, writePretty);
         });
     } else {
         var dir = path.dirname(outfile);
         mkdirp.sync(dir);
         data = fs.readFileSync(filepath, 'utf8');
-        makePretty(data, config, outfile, writePretty);
+        makePretty(data, config, filepath, outfile, writePretty);
     }
 }
 
-function makePretty(code, config, outfile, callback) {
+function makePretty(code, config, filepath, outfile, callback) {
     try {
         var fileType = getOutputType(outfile, config.type);
         var pretty = beautify[fileType](code, config);
 
-        callback(null, pretty, outfile, config);
+        callback(null, pretty, filepath, outfile, config, code);
     } catch (ex) {
         callback(ex);
     }
 }
 
-function writePretty(err, pretty, outfile, config) {
+function writePretty(err, pretty, filepath, outfile, config, code) {
     if (err) {
         console.error(err);
         process.exit(1);
     }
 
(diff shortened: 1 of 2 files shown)


bitwiseman:
Good start, thanks!  Now make it pass tests and also implement it python.

coderhaoxin:
ok. soon

coderhaoxin:
> also implement it python.

done

test seems failed on `master` too.
```
js-beautify output for /tmp/js-beautify-mkdir/example1-default.js was expected to be different
```

run the cli directly is ok, file created in `/tmp/js-beautify-mkdir`

bitwiseman:
No, the problem is the test expects even unchanged files to be written out...

js/lib/cli.js
```diff
@@ -293,24 +293,26 @@ function makePretty(code, config, outfile, callback) {
...
         process.exit(1);
     }
 
     if (outfile) {
-        try {
-            fs.writeFileSync(outfile, pretty, 'utf8');
-            logToStdout('beautified ' + path.relative(process.cwd(), outfile), config);
-        } catch (ex) {
-            onOutputError(ex);
+        if (code !== pretty) {
```
bitwiseman: Here's the problem.  You can't just not write the code if it is the same.  You need to check (around line 267) if the input and output path are the same.  If so, you can check here if the input and output code is unchanged and not change that file.  If the input path is empty, output path is empty (for piping cases), or if the input path differs from the output path you still have to write the output, even if it is unchanged.

js/lib/cli.js
```diff
@@ -278,39 +278,41 @@ function processInputSync(filepath) {
...
 }
 
-function makePretty(code, config, outfile, callback) {
+function makePretty(code, config, filepath, outfile, callback) {
     try {
         var fileType = getOutputType(outfile, config.type);
         var pretty = beautify[fileType](code, config);
 
-        callback(null, pretty, outfile, config);
+        callback(null, pretty, filepath, outfile, config, code);
```
bitwiseman:
```js
if (!(data === pretty && filepath === outfile)) {
    callback(null, pretty, outfile, config);
} 
```

Then you can revert the changes in `writePretty`.

bitwiseman:
Cool. We don't have a test to show that unchanged files are not written, but I can do that.  And all the tests pass.  Thanks!

coderhaoxin:
:smile:

blino:
The new tests seem to be OSX-specific, touch -A does not work with touch from GNU coreutils (Linux).

bitwiseman:
@blino - please open a new issue, and if you are willing to contribute a fix that would be even better!

Issue #598, opened 2014-12-26
Keep trailing spaces on comments


sitegui:
Hello,

I've noticed that trailing spaces on comments are stripped in some cases. Some examples that keep them:
```
// A line comment  
var a = 2
```
```
/*
A multiline with no star  
*/
var a = 2
```

But this one does not:
```
/*
* A multiline with star  
*/
var a = 2
```
(Tested on web)

For my use case, this is bad because js-docs comments are processed as Markdown, and two trailing spaces is the markdown way to print a new line in the final rendering.

This seems like a bug, since some formats do keep the trailing space, but I would like some confirmation on that.

Happy Holidays,
Gui

bitwiseman:
Interesting.  Shouldn't be hard to fix.

bitwiseman:
https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L1157
If we change that `trim` to `ltrim`, I think that would do it.

PR #620, opened 2015-02-06
Fixing the issue #598

websiddu:
Requested to review this commit, if you are ok with this I will port this to python also.


websiddu:
I am not sure why this happens!! Can any one help me ?

```bash
    Post-build git status check...
    Ensuring no changes visible to git have been made to '' ...
    ERROR: Post-build git status check - FAILED.
    Git status reported changes to non-git-ignore'd files.
    TO REPRO: Run 'git status '.  The use git gui or git diff to see what was changed during the build.
    TO FIX: Amend your commit and rebuild. Repeat until git status reports no changes both before and after the build.
    OUTPUT FOR 'git status ':
    # HEAD detached at FETCH_HEAD
    # Changes not staged for commit:
    #   (use "git add <file>..." to update what will be committed)
    #   (use "git checkout -- <file>..." to discard changes in working directory)
    #
    # modified:   js/test/beautify-javascript-tests.js
    #
    no changes added to commit (use "git add" and/or "git commit -a")
    make: *** [status] Error 1
    The command "make" exited with 2.
    Done. Your build exited with 1. 
```

bitwiseman:
Tests are generated during buld.  Do not edit `js/test/beautify-javascript-tests.js` directly.  Edit `test/data/javascript.js` instead.

js/lib/beautify.js
```diff
@@ -3,7 +3,7 @@
 
   The MIT License (MIT)
 
-  Copyright (c) 2007-2013 Einar Lielmanis and contributors.
+  Copyright (c) 2007-2015 Einar Lielmanis and contributors.
```
bitwiseman: If you must do this put it in another commit and do it everywhere.  It is unrelated to this commit.

js/lib/beautify.js
```diff
@@ -386,12 +386,13 @@
             var out = [],
```
bitwiseman: Update the comment on line 380, to talk about this change.

test/template/node-javascript.mustache
```diff
@@ -62,8 +62,8 @@ function run_javascript_tests(test_obj, Urlencoded, js_beautify, html_beautify,
         // }
 
         if (opts.indent_size === 4 && input) {
-            wrapped_input = '{\n' + input.replace(/^(.+)$/mg, '    $1') + '\n    foo = bar;\n}';
-            wrapped_expectation = '{\n' + expectation.replace(/^(.+)$/mg, '    $1') + '\n    foo = bar;\n}';
+            wrapped_input = '{\n' + input.trim().replace(/^(.+)$/mg, '    $1') + '\n    foo = bar;\n}';
```
bitwiseman: You really shouldn't put trim here...  It means your tests aren't testing what they intend to.

bitwiseman:
Hmm... #598 is caused by https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L1157
If you change that `trim` to `ltrim` and this fix the tests, I think that would do it.

websiddu:
For the above comment are you saying that the trailing spaces should be preserved ?

bitwiseman:
Yes, the bug is a little unclear, but it is a request to *not* strip trailing spaces in comments.

websiddu:
I don't think its a good practice to preserve the trailing spaces.. Anyways, I followed your suggestions and used `ltrim` in javadoc block which fixes the issue.. @bitwiseman

bitwiseman:
It isn't good practice to preserve trailing spaces.  We don't except in the case of comment where it makes sense to preserve them.  
This looks good.  Thanks for seeing it through.

PR #499, opened 2014-08-07
Add wrap_attributes and wrap_indent config parameters for HTML

sankargorthi:
This will put attributes on new lines if configured

#### input
-------
```html
<div    {{something}} attr1="123" data-attr2="hello    t here" {{#test}}disabled {{/test}}

>
This is some content    </div>
<div oneattr="one">
    Some content
            </div>
```

#### `config.json`
```json
{
    "wrap_attributes": true,
    "wrap_indent": 8,
    "wrap_line_length": 144
}
```

#### output
```html
<div {{something}}
        attr1="123"
        data-attr2="hello    t here"
        {{#test}}disabled
        {{/test}}>
    This is some content</div>
<div oneattr="one">
    Some content
</div>

```

PR #625, opened 2015-02-22
Wrap/attributes

filipesilva:
Picked up on #499, merged master, added support for short closed tags, added tests.


filipesilva:
Huh, I didn't put the new tests in the right place actually. I'll move them to the template file.

sankargorthi:
This is great! Thank you!

filipesilva:
No, thank YOU :p

I just tried to make your code run on master really, I don't understand much about how jsbeautify works on the inside.

bitwiseman:
Hmm... Could you squash your changes into one commit (or at least a smaller number)?

bitwiseman:
@filipesilva - I don't have to time to look it this closely yet, but it looks like you mostly did the right thing. 
Please remove your changes from the mustache file, and run `make all`. When that is done, you'll see that your changes in the html test data file will have propagated to the html test file.  Sorry for the lack of documentation.

filipesilva:
I can squash it, sure. I'm working through this tests data, it's giving me some trouble but I think I'm getting the idea.

filipesilva:
Ok seems to be done.

filipesilva:
I've reverted the file that didn't belong in the PR and squashed the commits again. I'm sorry if I'm being noobish, but I am new at this.

sankargorthi:
Will try and test this locally as well. Thanks so much for this.

bitwiseman:
@filipesilva - You're doing great, I'm impressed with how quickly you're working your way through it.

js/lib/beautify-html.js
```diff
@@ -362,7 +377,7 @@
                     }
 
                     if (indent_handlebars && !tag_start_char) {
-                        if (content.length >= 2 && content[content.length - 1] === '{' && content[content.length - 2] == '{') {
+                        if (content.length >= 2 && content[content.length - 1] === '{' && content[content.length - 2] === '{') {
```
bitwiseman: I've made this change and the extra space clean up in the master branch now.  Thanks for bringing them up.

README.md
```diff
@@ -98,6 +98,8 @@ Beautifier Options:
   -k, --keep-array-indentation      Preserve array indentation
   -x, --unescape-strings            Decode printable characters encoded in xNN notation
   -w, --wrap-line-length            Wrap lines at next opportunity after N characters [0]
+  -a, --wrap-attributes             Wrap attributes to new lines
+  -i, --wrap-indent                 Indent wrapped attributes to after N characters [indent-size]
```
bitwiseman: Let's change the name of this to `wrap-attributes-indent-size`.  It makes it clean the the indent we're talking about is specific to the attributes.

README.md
```diff
@@ -98,6 +98,8 @@ Beautifier Options:
   -k, --keep-array-indentation      Preserve array indentation
   -x, --unescape-strings            Decode printable characters encoded in xNN notation
   -w, --wrap-line-length            Wrap lines at next opportunity after N characters [0]
+  -a, --wrap-attributes             Wrap attributes to new lines
```
bitwiseman: `-a` is already used by the js-beautify component.  How about `-A` and `-I`?

bitwiseman: Instead of a true/false, please make this a string valid options for now are "auto" (default) and "force".  This will leave it open for us to add "none" or other option later.

bitwiseman:
@filipesilva - You're doing great, I'm impressed with how quickly you're working your way through it.  

Sorry, for the multiple iterations to find the issues.  We're getting close.  Please  add a few more tests, including showing what happens when the attributes wrap due to long line length.

README.md
filipesilva: Will do.

README.md
filipesilva: I'll move them down and change the -a to -W, I think it makes more sense than -A if it's colliding with other options since it is related to wrap (-w). Also looking into the options.

filipesilva: Just noticed that there is what seems to be an obsolete option using -W. I'll change the wrap-attributes to -A so there's no collision, since removing the other option shouldn't be done on this PR.

filipesilva:
Alright, I think that commit encompasses the changes you mentioned.

README.md
```diff
@@ -122,7 +122,9 @@ These largely correspond to the underscored option keys for both library interfa
     "break_chained_methods": false,
     "eval_code": false,
     "unescape_strings": false,
-    "wrap_line_length": 0
+    "wrap_line_length": 0,
+    "wrap_attributes": "auto",
+    "wrap_attributes_indent_size": 4
```
bitwiseman: This should default to `indent_size`.  For example, if `indent_size`is set to 2, and this is not set, it should default to 2.

js/lib/beautify-html.js
```diff
@@ -342,6 +347,16 @@
...
                         this.space_or_wrap(content);
                         space = false;
+                        if (!first_attr && wrap_attributes === 'force' &&  input_char !== '/') {
+                            this.print_newline(true, content);
+                            this.print_indentation(content);
+                            for (var count = 0; count < wrap_attributes_indent_size; count++) {
+                                content.push(indent_character);
+                            }
+                        }
+                        if (content.filter(function (i) { return i === ' ';}).length === 1){
```
bitwiseman: This is terse and cool, but is not compatible with old browsers.  Could you do this as for loop?

test/data/html.js
```diff
@@ -31,6 +31,53 @@ exports.test_data = {
...
+              eof: '\\n',
+              indent_attr: '        '
+          }, {
+              options: [
+                  { name: "wrap_attributes", value: "'auto'" },
+                  { name: "wrap_line_length", value: "0" }
+              ],
+              eof: ' ',
+              indent_attr: ''
+          }
```
bitwiseman: Add a matrix entry for:
```js
              options: [
                  { name: "wrap_attributes", value: "'auto'" },
                  { name: "wrap_line_length", value: "80" }
              ],
```

bitwiseman:
Couple more little tweaks.

filipesilva:
Done and done.

bitwiseman:
Like a boss.

filipesilva:
Yeahhhhhhhhh :D

darkbls:
nice work

PR #633, opened 2015-03-06
Add GUI support for `--indent-inner-html`.

bardiharborow:
Creates a GUI option to go from:
```html
<html>

<head></head>

<body></body>

</html>
```
To:
```html
<html>

    <head></head>

    <body></body>

</html>
```
I would like to also add an option to go to:
```html
<html>
    <head></head>
    <body></body>
</html>
```
But maybe later, seeing as the underlying engine does not seem to support it (?).


bitwiseman:
Nice job.  Looks good.

bitwiseman:
Please file an issue for the the second case you referred to.

PR #636, opened 2015-03-07
Issue 635 - leave handelbars comments as unformatted

donwalling:
This is a code suggestion for issue 635 to leave handelbars comments unformatted like html comments. I haven't figured out the test case heirarchy yet to understand where to put a test case. Suggestions would be appreciated.

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -405,7 +405,7 @@
                     if (indent_handlebars && content[1] && content[1] === '{' && content[2] && content[2] === '!') { //if we're in a comment, do something special
                         // We treat all comments as literals, even more than preformatted tags
                         // we just look for the appropriate close tag
-                        content = [this.get_handlebars_comment(tag_start)];
+                        content = [this.get_comment(tag_start)];
                         break;
                     }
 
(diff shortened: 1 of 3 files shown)


bitwiseman:
Haven't reviewed, but at least need tests.

bitwiseman:
@donwalling - we're going to need tests for this before I can review it.

donwalling:
Hi @bitwiseman - does https://github.com/donwalling/js-beautify/commit/2bcf4153748389cbca6dfd7f6787c3bdab26e841 suffice for tests? Basically re-using what is already there in a second variation. The challenge I ran into with other approaches is that mustache is stripping out the {{! }} comments when the script creates the beautify-html-tests.js file. 

BTW, I'm not sure why the beautify-html-tests.js file is checked in? it seems to be created by the testj Makefile target

js/lib/beautify-html.js
```diff
@@ -536,6 +545,31 @@
                 return comment;
             };
 
+            this.get_handlebars_comment = function(start_pos) { //function to return handlebars comment content in its entirety
```
bitwiseman: Any reason why this wouldn't be part of the general comment method above? it's mostly the same...

bitwiseman: Shouldn't this be added to the comment handling method above instead of being made into a separate method?

donwalling: indeed, it should have been - will submit another commit

bitwiseman:
Your commit won't merge right now because I went in and made the handlebars tests part of the standard test template system.

bitwiseman:
I rebased your changes to a single commit and then added the tests in the new format.   Thanks!

PR #672, opened 2015-04-19
Add consistent_comment_rule option for CSS

leoliu1313:
consistent_comment_rule option is false by default.
This change is not only in javascript, but also in python codes and regressions.
This shall be an option to be used for brackets-beautify project as an extension for Brackets.
https://github.com/leoliu1313/brackets-beautify
https://github.com/drewhamlett/brackets-beautify
http://brackets.io/


leoliu1313:
Don't know how to get rid of git-status-clear.sh check. Revert all the changes and try again.

Issue #645, opened 2015-03-10
CSS comment spacing disregards `newline_between_rules`, `selector_separator_newline`


jawshooah:
Original:

``` css
/* Comment above first rule */

body {
    display: none;
}

/* Comment between rules */

ul,

/* Comment between selectors */

li {
    display: none;
}

/* Comment after last rule */
```

Regardless of the values of `newline_between_rules` and `selector_separator_newline`, the output is the same:

``` css
/* Comment above first rule */

body {
    display: none;
}
/* Comment between rules */

ul,
/* Comment between selectors */

li {
    display: none;
}
/* Comment after last rule */
```

anephew:
Fixed?

bitwiseman:
Still open, so not fixed.

MacKLess:
This appears to have been resolved as of v1.8.0-rc13. @jawshooah  and @anephew, if this doesn't appear to be working correctly, can you please let me know.

Issue #609, opened 2015-01-27
css beautify adding an extra newline after a comment line in a css block


MetaMemoryT:
original:

```
.row {
  /* margin: -(gutters.horizontal) 0 -1px -(gutters.vertical) */
  margin: -0px 0 -1px -50px !important;
}
```

formatted:
```
.row {
  /* margin: -(gutters.horizontal) 0 -1px -(gutters.vertical) */

  margin: -0px 0 -1px -50px !important;
}
```

What is expected: same as original.

Issue #531, opened 2014-09-18
Extra newline is inserted after the comment line instead of before it


archersunx:
Please have a check on below css codes,
**Original codes:**

``` css
/* Space out content a bit */
body {
  padding-top: 20px;
  padding-bottom: 20px;
}

/* Custom page header */
.header {
  border-bottom: 1px solid #e5e5e5;
}
```

**After formatted:**

``` css
/* Space out content a bit */

body {
    padding-top: 20px;
    padding-bottom: 20px;
}
/* Custom page header */

.header {
    border-bottom: 1px solid #e5e5e5;
}
```

Extra newline is inserted after the comment line instead of before it. I checked out the source code, it seems like a intentional behavior, but it sounds wired for me, as the comment is obviously for the next rule set, not the previous one.

bor:
I've catched the same issue

bitwiseman:
Too busy to work on this at this time.  Please help by submitting a PR with fix and tests.

robsonrosa:
Same here. But adding a few more details, it  behaves differently depending on how you add your comment. 

`block comment` **outside** brackets adds an extra line before and after the comment each time you format.
`block comment` **inside** brackets works fine.
`inline comment` **outside** brackets works fine.
`inline comment` **inside** removes all spaces between comment and last character.

Example before formatting:

![image](https://cloud.githubusercontent.com/assets/2868031/26420810/31054692-409a-11e7-8e6d-040507b239a0.png)

Example after formatting:
![image](https://cloud.githubusercontent.com/assets/2868031/26420985/c22def5c-409a-11e7-9d81-58c4287e4274.png)

It's weird. If you use `inline comment` **outside** brackets and `block comment` **inside** brackets, everything works fine.

NathanStrutz:
Please note that CSS does not have inline comments. Have we thought about the correct, most beautiful way to reformat this:

```css
/* block comment */
.selector {
  width: 0; /* inline comment */
}
```

It should either leave it alone, or it should move the inline comment above the line it comments on. I would never want to see the inline comment on its own line after the one it's marking up, which is what it does with the current release, afaik.

PR #679, opened 2015-04-27
Add consistent_comment_rule option for CSS

leoliu1313:


leoliu1313:
consistent_comment_rule option is false by default.
This change is not only in javascript, but also in python codes and regressions.
This shall be an option to be used for brackets-beautify project as an extension for Brackets.
https://github.com/leoliu1313/brackets-beautify
https://github.com/drewhamlett/brackets-beautify
http://brackets.io/

leoliu1313:
Travis CI is clean now.
https://travis-ci.org/beautify-web/js-beautify/builds/60225208
https://travis-ci.org/leoliu1313/js-beautify

Travis CI error is fixed by uploading the output files generated by Makefile.

bitwiseman:
Excellent.  Yes, checking in the outputs in this case is correct. I'll review this shortly.

leoliu1313:
Thank you!

leoliu1313:
This shall fix the folloing issues.
https://github.com/beautify-web/js-beautify/issues/531
https://github.com/beautify-web/js-beautify/issues/609
https://github.com/beautify-web/js-beautify/issues/645

loganknecht:
Hi there, I'm interested to know if this has been reviewed and if its integration is planned.

Specifically because I'm looking for a solution to fix the issue I'm encountering with #609.

bitwiseman:
Reviewed now.  I'm not convinced this is the right path for this.  This doesn't so much fix the issues you listed as create a different set of issues.  If we're going to make this change it shouldn't be an option, it should be the standard behavior.   

Having "consistent" behavior is good, but this implementation produces worse outputs in some cases.

bitwiseman:
@loganknecht - #609 fixed.  Will be in next release coming within the week.

PR #683, opened 2015-05-01
Don't wrap XML declaration (or processing instructions)

moreginger:
Don't wrap XML declaration (or processing instructions) if wrap_attributes='force'.

Tried to add tests for this but they fail. Possibly due to tag type?


bitwiseman:
Not sure, but if you run "make" locally, you should see the same errors. 
NOTE: you have to check in the file generated from the change to the test data file.

Issue #587, opened 2014-11-30
No option to "Indent with tabs" for HTML files


whatsnewsaes:
Indent_with_tabs: true ---- this option is only available under the "JS" section.

I tried copying this rule to the HTML section. Did not work

rlindner81:
I was looking for this today and it is indeed missing. 

The available options for both HTML and CSS are fewer than for JS, as stated in
https://github.com/beautify-web/js-beautify/blob/master/README.md

In my opinion, it would make a lot of sense to at least add the tabs option.

PR #684, opened 2015-05-03
Fixes issue #587 for javascript implementation

rlindner81:
Simple fix to interpret ident-with-tabs option in javascript implementation

### DIFF
--- a/js/lib/beautify-css.js
+++ b/js/lib/beautify-css.js
@@ -74,6 +74,10 @@
             indentSize = parseInt(indentSize, 10);
         }
 
+        if(options.indent_with_tabs){
+            indentCharacter = '\t';
+            indentSize = 1;
+        }
 
         // tokenizer
         var whiteRe = /^\s+$/;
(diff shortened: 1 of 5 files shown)


bitwiseman:
@rlindner81 - execellent!  Please add the python implementation for the css (it is easy, trust me) and add some tests to https://github.com/beautify-web/js-beautify/tree/master/test/data.  (The change to the data files will result in change to generated test files.  Check those in as well.)

rlindner81:
I added the python implementation for CSS (it was easy). I did not see any tests for the indent_with_tabs option (even for js). Can you point them out?

bitwiseman:
Looks like your python needs a tweak.   Before you push, run `make` from root if you want to test your changes. (This will report failure if you have files checked out.) 
There might now be any tests for this option.  Look at 
https://github.com/beautify-web/js-beautify/blob/master/test/data/javascript.js#L48 for an example of what a test section might look like.  But yours can be much shorter and simpler. :smile: 

Here is an example of a simple change: 
https://github.com/beautify-web/js-beautify/commit/46cf32151a6378814a13d3d8c925e1c649ba41ac

https://github.com/beautify-web/js-beautify/commit/46cf32151a6378814a13d3d8c925e1c649ba41ac

to see what a basic option test looks like.

rlindner81:
Well that was fun.

rlindner81:
@bitwiseman Still unhappy?

kudou:
@bitwiseman STILL NOT WORK

rlindner81:
@kudou Please avoid all caps. And be explicit with what you mean. What isn't working?

PR #693, opened 2015-05-17
Fix tokenizer's bracket pairs' open stack

redapple:
I found this bug while parsing things like:
```
for (index in [1,2,3]) /^salign$/i.test('salign')
```
(from https://github.com/rspivak/slimit/issues/54)

Before the fix, you get:
```
for (index in [1, 2, 3]) / ^ salign$ / i.test('salign')
```
The regular expression is not detected.

With this fix, the regular expression literal is kept as-is:
```
>>> import jsbeautifier
>>> print jsbeautifier.beautify("for (index in [1,2,3]) /^salign$/i.test('salign')")
for (index in [1,2,3]) /^salign$/i.test('salign')
>>> 
```

### DIFF
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -1518,13 +1518,13 @@
 
                 if (next.type === 'TK_START_BLOCK' || next.type === 'TK_START_EXPR') {
                     next.parent = last;
+                    open_stack.push(open);
                     open = next;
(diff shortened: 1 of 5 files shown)


bitwiseman:
Strange symptom.  Good catch on the fix though, thanks!

PR #696, opened 2015-05-22
fix global variable result issue

wh1100717:


bitwiseman:
is there a test you can write that shows the failure that this fixes?

bitwiseman:
Nevermind.  General correctness.  Thanks!

PR #711, opened 2015-06-11
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "js-beautify",
-  "version": "1.5.6",
   "main": [
     "./js/lib/beautify.js",
     "./js/lib/beautify-css.js",


bitwiseman:
It says "moot at the moment".  It costs almost nothing to keep this updated right now.  Is there a real win for removing it?

kkirsche:
Also just an FYI there are discussions to move bower to NPM anyway so I wouldn't really plan to support it for ever: bower/bower#1520. I think it's a bad decision to encourage bower / keep unused properties

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

bitwiseman:
Thanks! One less thing I need to do for releases.

PR #730, opened 2015-07-08
Remove unnecessary legacy tests

olsonpm:
Since the legacy and new names point to the same object,
  there is no need to run through all the tests separately


bitwiseman:
Looks good.  Thanks!

PR #735, opened 2015-07-16
first draft of operator_position functionality

olsonpm:
- Added verification of operator_position option.  Throws error if
    invalid.
 - Possible operator_position options are:
    1) undefined (use default behavior)
    2) 'before_newline'
    3) 'after_newline'
    4) 'preserve_newline'
    5) 'remove_newline'
 - Moved some code around in handle_operator in order to separate
    conflicting logic when handling operators.
 - Modified allow_wrap_or_preserved_newline to handle operator
    preserved newlines as well.
 - Fixed a few boolean variable assignments within if statements.
    Though not always the case, the expression within the if can
    be assigned directly to the variable
 - Added -> built tests
 - Within the tests, added an 'inputlib.js' file which holds common
    input arrays.  This was necessary due to the matrices not working
    cleanly with default behavior.
 - It is worth noting that although the python test file was modified,
    this is a result of the built test and not actually providing
    a python port for this option (yet).

### DIFF
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -84,6 +84,23 @@
 
 */
 
+// Object.values polyfill found here:
+// http://tokenposts.blogspot.com.au/2012/04/javascript-objectkeys-browser.html
+if (!Object.values) {
+    Object.values = function(o) {
+        if (o !== Object(o)) {
+          throw new TypeError('Object.values called on a non-object');
+        }
+        var k=[],p;
+        for (p in o) {
+          if (Object.prototype.hasOwnProperty.call(o,p)) {
+            k.push(o[p]);
+          }
+        }
+        return k;
+    };
+}
+
 (function() {
 
     var acorn = {};
@@ -178,6 +195,28 @@
         var beautifier = new Beautifier(js_source_text, options);
         return beautifier.beautify();
     }
+
+    function sanitizeOperatorPosition(opPosition) {
+        opPosition = opPosition || OPERATOR_POSITION.before_newline;
+
+        var validPositionValues = Object.values(OPERATOR_POSITION);
+
+        if (!in_array(opPosition, validPositionValues)) {
+            throw new Error("Invalid Option Value: The option 'operator_position' must be one of the following values\n"
+                + validPositionValues
+                + "\nYou passed in: '" + opPosition + "'");
+        }
+
+        return opPosition;
+    }
+
+    var OPERATOR_POSITION = {
+        before_newline: 'before-newline',
+        after_newline: 'after-newline',
+        preserve_newline: 'preserve-newline',
+    };
+
+    var OPERATOR_POSITION_BEFORE_OR_PRESERVE = [OPERATOR_POSITION.before_newline, OPERATOR_POSITION.preserve_newline];
 
     var MODE = {
             BlockStatement: 'BlockStatement', // 'BLOCK'
@@ -288,6 +327,7 @@
         opt.e4x = (options.e4x === undefined) ? false : options.e4x;
         opt.end_with_newline = (options.end_with_newline === undefined) ? false : options.end_with_newline;
         opt.comma_first = (options.comma_first === undefined) ? false : options.comma_first;
+        opt.operator_position = sanitizeOperatorPosition(options.operator_position);
 
         // For testing of beautify ignore:start directive
         opt.test_output_raw = (options.test_output_raw === undefined) ? false : options.test_output_raw;
@@ -440,7 +480,19 @@
                 return
             }
 
-            if ((opt.preserve_newlines && current_token.wanted_newline) || force_linewrap) {
+            var shouldPreserveOrForce = (opt.preserve_newlines && current_token.wanted_newline) || force_linewrap;
+            var operatorLogicApplies = in_array(flags.last_text, Tokenizer.positionable_operators) || in_array(current_token.text, Tokenizer.positionable_operators);
+
+            if (operatorLogicApplies) {
+              var shouldPrintOperatorNewline = (
+                  in_array(flags.last_text, Tokenizer.positionable_operators) &&
+                  in_array(opt.operator_position, OPERATOR_POSITION_BEFORE_OR_PRESERVE)
+                )
+                || in_array(current_token.text, Tokenizer.positionable_operators);
+              shouldPreserveOrForce = shouldPreserveOrForce && shouldPrintOperatorNewline;
+            }
+
+            if (shouldPreserveOrForce) {
                 print_newline(false, true);
             } else if (opt.wrap_line_length) {
                 if (last_type === 'TK_RESERVED' && in_array(flags.last_text, newline_restricted_tokens)) {
@@ -1185,6 +1237,18 @@
                 return;
             }
 
+            if (current_token.text === '::') {
+                // no spaces around exotic namespacing syntax operator
+                print_token();
+                return;
+            }
+
+            // Allow line wrapping between operators when operator_position is
+            //   set to before or preserve
+            if (last_type === 'TK_OPERATOR' && in_array(opt.operator_position, OPERATOR_POSITION_BEFORE_OR_PRESERVE)) {
+                allow_wrap_or_preserved_newline();
+            }
+
             if (current_token.text === ':' && flags.in_case) {
                 flags.case_body = true;
                 indent();
@@ -1194,21 +1258,86 @@
                 return;
             }
 
-            if (current_token.text === '::') {
-                // no spaces around exotic namespacing syntax operator
-                print_token();
-                return;
-            }
-
-            // Allow line wrapping between operators
-            if (last_type === 'TK_OPERATOR') {
-                allow_wrap_or_preserved_newline();
-            }
-
             var space_before = true;
             var space_after = true;
+            var in_ternary = false;
+            var isGeneratorAsterisk = current_token.text === '*' && last_type === 'TK_RESERVED' && flags.last_text === 'function';
+            var isUnary = in_array(current_token.text, ['-', '+']) && (
+                in_array(last_type, ['TK_START_BLOCK', 'TK_START_EXPR', 'TK_EQUALS', 'TK_OPERATOR']) ||
+                in_array(flags.last_text, Tokenizer.line_starters) ||
+                flags.last_text === ','
+              );
 
-            if (in_array(current_token.text, ['--', '++', '!', '~']) || (in_array(current_token.text, ['-', '+']) && (in_array(last_type, ['TK_START_BLOCK', 'TK_START_EXPR', 'TK_EQUALS', 'TK_OPERATOR']) || in_array(flags.last_text, Tokenizer.line_starters) || flags.last_text === ','))) {
+            if (current_token.text === ':') {
+                if (flags.ternary_depth === 0) {
+                    // Colon is invalid javascript outside of ternary and object, but do our best to guess what was meant.
+                    space_before = false;
+                } else {
+                    flags.ternary_depth -= 1;
+                    in_ternary = true;
+                }
+            } else if (current_token.text === '?') {
+                flags.ternary_depth += 1;
+            }
+
+            // let's handle the operator_position option prior to any conflicting logic
+            if (!isUnary && !isGeneratorAsterisk && opt.preserve_newlines && in_array(current_token.text, Tokenizer.positionable_operators)) {
+                var isColon = current_token.text === ':';
+                var isTernaryColon = (isColon && in_ternary);
+                var isOtherColon = (isColon && !in_ternary);
+
+                switch (opt.operator_position) {
+                    case OPERATOR_POSITION.before_newline:
+                        // if the current token is : and it's not a ternary statement then we set space_before to false
+                        output.space_before_token = !isOtherColon;
+
+                        print_token();
+
+                        if (!isColon || isTernaryColon) {
+                            allow_wrap_or_preserved_newline();
+                        }
+
+                        output.space_before_token = true;
+                        return;
+
+                    case OPERATOR_POSITION.after_newline:
+                        // if the current token is anything but colon, or (via deduction) it's a colon and in a ternary statement,
+                        //   then print a newline.
+
+                        output.space_before_token = true;
+
+                        if (!isColon || isTernaryColon) {
+                            if (get_token(1).wanted_newline) {
+                                print_newline(false, true);
+                            } else {
+                                allow_wrap_or_preserved_newline();
+                            }
+                        } else {
+                            output.space_before_token = false;
+                        }
+
+                        print_token();
+
+                        output.space_before_token = true;
+                        return;
+
+                    case OPERATOR_POSITION.preserve_newline:
+                        if (!isOtherColon) {
+                            allow_wrap_or_preserved_newline();
+                        }
+
+                        // if we just added a newline, or the current token is : and it's not a ternary statement,
+                        //   then we set space_before to false
+                        space_before = !(output.just_added_newline() || isOtherColon);
+
+                        output.space_before_token = space_before;
+                        print_token();
+                        output.space_before_token = true;
+                        return;
+                }
+            }
+
+            if (in_array(current_token.text, ['--', '++', '!', '~']) || isUnary) {
                 // unary operators (and binary +/- pretending to be unary) special cases
 
                 space_before = false;
@@ -1250,16 +1379,7 @@
                     // foo(); --bar;
                     print_newline();
                 }
-            } else if (current_token.text === ':') {
-                if (flags.ternary_depth === 0) {
-                    // Colon is invalid javascript outside of ternary and object, but do our best to guess what was meant.
-                    space_before = false;
-                } else {
-                    flags.ternary_depth -= 1;
-                }
-            } else if (current_token.text === '?') {
-                flags.ternary_depth += 1;
-            } else if (current_token.text === '*' && last_type === 'TK_RESERVED' && flags.last_text === 'function') {
+            } else if (isGeneratorAsterisk) {
                 space_before = false;
                 space_after = false;
             }
@@ -1273,9 +1393,7 @@
                 output.add_raw_token(current_token)
                 if (current_token.directives && current_token.directives['preserve'] === 'end') {
                     // If we're testing the raw output behavior, do not allow a directive to turn it off.
-                    if (!opt.test_output_raw) {
-                        output.raw = false;
-                    }
+                    output.raw = opt.test_output_raw;
                 }
                 return;
             }
@@ -1308,12 +1426,8 @@
             // block comment starts with a new line
             print_newline(false, true);
             if (lines.length > 1) {
-                if (all_lines_start_with(lines.slice(1), '*')) {
-                    javadoc = true;
-                }
-                else if (each_line_matches_indent(lines.slice(1), lastIndent)) {
-                    starless = true;
-                }
+                javadoc = all_lines_start_with(lines.slice(1), '*');
+                starless = each_line_matches_indent(lines.slice(1), lastIndent);
             }
 
             // first line always indented
@@ -1613,7 +1727,11 @@
         var digit_oct = /[01234567]/;
         var digit_hex = /[0123456789abcdefABCDEF]/;
 
-        var punct = ('+ - * / % & ++ -- = += -= *= /= %= == === != !== > < >= <= >> << >>> >>>= >>= <<= && &= | || ! ~ , : ? ^ ^= |= :: => **').split(' ');
+        this.positionable_operators = '!= !== % & && * ** + - / : < << <= == === > >= >> >>> ? ^ | ||'.split(' ')
+        var punct = this.positionable_operators.concat(
+            // non-positionable operators - these do not follow operator position settings
+            '! %= &= *= ++ += , -- /= :: <<= = => >>= >>>= ^= |= ~'.split(' '))
+
         // words which should always start on new line.
         this.line_starters = 'continue,try,throw,return,var,let,const,if,switch,case,default,for,while,break,function,import,export'.split(',');
         var reserved_words = this.line_starters.concat(['do', 'in', 'else', 'get', 'set', 'new', 'catch', 'finally', 'typeof', 'yield', 'async', 'await', 'from', 'as']);
@@ -2177,7 +2295,6 @@
             }
             return out;
         }
-
     }
 
 
--- a/test/data/javascript/tests.js
+++ b/test/data/javascript/tests.js
@@ -1,3 +1,5 @@
+var inputlib = require('./inputlib');
+
 exports.test_data = {
     default_options: [
         { name: "indent_size", value: "4" },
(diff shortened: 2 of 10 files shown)


olsonpm:
So instead of figuring out the redundant indentation logic and attempt to condense the tests via matrices, I just made it easy on myself and separated the tests.  I added tests for each combination of operator_position and preserve_newlines, meaning there should be 10 new sections of tests total.

Also, as noted in the commit message - I'm expecting the python tests to fail until we verify the javascript functionality.  I can write the logic for python afterward.

bitwiseman:
Let me start by saying - great job.  Thanks for putting the full effort into this and I approve of everything you mentioned in your description, including not using matrix style for testing.  

The size of the change is a little daunting - though I get that the test portions are largely boilerplate. 

I'll take a look at the overall structure of the product code changes so that you are unblocked to do the python porting.  Then I'll look at the detailed tests to make sure everything is coming out the way it should.

Thanks again!

js/lib/beautify.js
```diff
@@ -279,6 +300,10 @@
         opt.e4x = (options.e4x === undefined) ? false : options.e4x;
         opt.end_with_newline = (options.end_with_newline === undefined) ? false : options.end_with_newline;
         opt.comma_first = (options.comma_first === undefined) ? false : options.comma_first;
+        opt.operator_position = options.operator_position;
+        
+        // throws error if invalid value was passed in
+        verifyOperatorPosition(opt.operator_position);
```
bitwiseman: Maybe change this to `sanitizeOperatorPosition` and have the function return the value to be assigned (replacing undefined with the default).

olsonpm:
Yeah I thought the testing might have been overkill.  It was more my method of development than me testing the end result (start by ensuring all the operators worked properly -> the special colon case -> then a catch all).

Feel free to let me know what you want slimmed down and I'll do it.  However I will wait to hear the full product-changes feedback before working on it further.

Finally, this is my first sizable open source contribution - so don't hesitate to critique anything.  Normally I just fork and modify for my personal needs.

js/lib/beautify.js
```diff
@@ -170,6 +170,27 @@
         var beautifier = new Beautifier(js_source_text, options);
         return beautifier.beautify();
     }
+    
+    function verifyOperatorPosition(opPosition) {
+        var validPositionValues = [undefined];
+        Object.keys(OPERATOR_POSITION).forEach(function(pos) {
```
bitwiseman: Not sure if this is compatible with older browsers  (IE is a pain). 

Here's a polyfill to work around it:
http://tokenposts.blogspot.com.au/2012/04/javascript-objectkeys-browser.html

But, you might have to do this instead:
```js 
opPosition = opPosition || OPERATOR_POSITION. before_newline;
if (!OPERATOR_POSITION[opPosition]) {
    throw new Error();
}
return opPosition;
```

js/lib/beautify.js
```diff
@@ -1115,11 +1150,87 @@
...
+                } else {
+                    flags.ternary_depth -= 1;
+                    in_ternary = true;
+                }
+            } else if (current_token.text === '?') {
+                flags.ternary_depth += 1;
+            }
+            
+            // let's handle the operator_position option prior to any conflicting logic
+            if (typeof opt.operator_position !== 'undefined' && in_array(current_token.text, OP_POSITION_OPERATORS)) {
```
bitwiseman: Handling the defaulting in the opt setting above means you don't have to check for undefined here.

js/lib/beautify.js
```diff
@@ -1193,9 +1292,7 @@
                 output.add_raw_token(current_token)
                 if (current_token.directives && current_token.directives['preserve'] === 'end') {
                     // If we're testing the raw output behavior, do not allow a directive to turn it off.
-                    if (!opt.test_output_raw) {
-                        output.raw = false;
-                    }
+                    output.raw = opt.test_output_raw;
```
bitwiseman: Nice.

js/lib/beautify.js
```diff
@@ -1228,12 +1325,8 @@
...
             print_newline(false, true);
             if (lines.length > 1) {
-                if (all_lines_start_with(lines.slice(1), '*')) {
-                    javadoc = true;
-                }
-                else if (each_line_matches_indent(lines.slice(1), lastIndent)) {
-                    starless = true;
-                }
+                javadoc = all_lines_start_with(lines.slice(1), '*');
+                starless = each_line_matches_indent(lines.slice(1), lastIndent);
```
bitwiseman: Nice.

test/data/javascript.js
```diff
@@ -92,6 +94,800 @@ exports.test_data = {
...
+        ],
+        tests: [
+            {
+                comment: 'comprehensive, but simple',
+                unchanged: inputlib.operator_position.comprehensive_simple
+            }, {
+                comment: 'comprehensive, various newlines',
+                input: inputlib.operator_position.comprehensive_newlines,
+                output: [
+                    'var res = a + b - c /',
```
bitwiseman: This seems wrong.  The default is not `remove_newline`, so if someone sets `operator_position  = remove_newline`, it should take precedence over the general `preserve_newlines = true` setting.   That is what you have it doing for `operator_position  = preserve_newline` with ``preserve_newlines = false`, so the inverse should also hold.

test/data/javascript.js
```diff
@@ -92,6 +94,800 @@ exports.test_data = {
...
+                    '} else if (!(complex && simple) || (emotion && emotion.name === "happy")) {',
+                    '    cryTearsOfJoy(many || anOcean || aRiver);',
+                    '}'
+                ]
+            }
+        ]
+    }, {
+        name: "operator_position option - set to default (undefined)",
+        options: [
+            { name: "operator_position", value: "undefined" },
```
bitwiseman: You know, this would be one spot where the matrix structure would help you.  You could clearly show that your testing for `operator_position = undefined` to have the same behavior as the default value.

js/lib/beautify.js
```diff
@@ -423,7 +448,11 @@
                 return
             }
 
-            if ((opt.preserve_newlines && current_token.wanted_newline) || force_linewrap) {
+            var shouldPreserveOperatorNewline = (opt.operator_position === OPERATOR_POSITION.preserve_newline) &&
+                in_array(flags.last_text, OP_POSITION_OPERATORS) ||
+                in_array(current_token.text, OP_POSITION_OPERATORS);
+                
+            if (((opt.preserve_newlines || shouldPreserveOperatorNewline) && current_token.wanted_newline) || force_linewrap) {
```
bitwiseman: (overide_general_preserve) 
This seems wrong.  The default is not `remove_newline`, so if someone sets `operator_position  = remove_newline`, it should take precedence over the general `preserve_newlines = true` setting.   That is what you have it doing for `operator_position  = preserve_newline` with ``preserve_newlines = false`, so the inverse should also hold.

bitwiseman: ... Actually, the more I think about it, if `preserve_newlines = false`, that setting should take precedence for input requested newlines.  That will make the behavior most consistent.  If you turn off preserve_newlines you just turn them _off_.  If you allow them, then you can control where they are (or remove them) for operators.

The line-wrapping behavior is covered in by when the allow_wrap_or_preserve_newline gets called...  I'll need to review the inputs and outputs more carefully for this case.

bitwiseman:
Okay, I'm done with my initial pass.  
There's the design issue of needing to make the `preserve_newlines <--> operator_position` interaction consistent, and some code tweaks, but rather minimal. 

Thanks again!

olsonpm:
Sounds good.  I'll have time this weekend to take a look and make changes.

js/lib/beautify.js
olsonpm: Can you clarify "replacing undefined with the default"?  I'm expecting undefined to be the default value.

And will do.

js/lib/beautify.js
olsonpm: Ha - I had no idea IE had trouble with that.  I'll implement it without the polyfill.

js/lib/beautify.js
olsonpm: To clarify my intent - I was assuming the default setting would be 'undefined', and that this library would handle operators as it always had.  That is why I'm checking for undefined, and skipping the logic if it is.

js/lib/beautify.js
olsonpm: I will wait to hear our resolution surrounding default behavior before discussing your above comments.

olsonpm:
Wow - I think I see my misunderstanding.  The operator_position shouldn't do anything if preserve_newlines is false.  It's only positioning the operators that are placed next to a newline that the user wants preserved.  This is why you're mentioning the default behavior 'undefined' should equal 'before_newline'.

I took the option names far too literally.  I'll work on this now and have an update tonight hopefully.

bitwiseman:
@olsonpm - I've been away. It looks like you've been away too.  You wan to take another swing at this?

olsonpm:
Not a problem.

I think where our ideas differ on how to handle default behavior (i.e. no option specified)

My understanding is that it should behave as it did before

It seems from your comment above that you before_newline to be the default behavior which will be a breaking change.

bitwiseman:
Hm, It's probably an error in understanding on my part.  
I thought `OPERATOR_POSITION.before_newline` is the same as the current behavior.   I'll go look again.

olsonpm:
Referring back to your sample code you can see a few subtle differences:

```javascript
// Test code
var d = 1;
if (a === b 
    && c) {
    d = ((c * everything / something_else)  
            - a) %
        b;
    e 
        += d;

} else if (!(complex && simple) ||
        (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many ||
        anOcean ||
        aRiver);
}

// Output: actual current 
var d = 1;
if (a === b && c) {
    d = ((c * everything / something_else) - a) %
        b;
    e
        += d;

} else if (!(complex && simple) ||
    (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many ||
        anOcean ||
        aRiver);
}

// Output: operator_positioning = before_newline
var d = 1;
if (a === b &&
    c) {
    d = ((c * everything / something_else) -
            a) %
        b;
    e +=
        d;

} else if (!(complex && simple) ||
        (emotion && emotion.strong && emotion.name === 'happy')) {
    cryTearsOfJoy(many ||
        anOcean ||
        aRiver);
}
```

bitwiseman:
Ah, I see where we're out of sync.  In all the complexity, I didn't cover the core cases. 

```js
  opts.preserve_newlines = false;
  opts.operator_position = undefined;
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 
  bt('var res = a + b\n' +
      '    - c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 

  opts.operator_position = 'before_newline';
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 
  bt('var res = a + b\n' +
      '    - c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 

  opts.operator_position = 'after_newline';
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 
  bt('var res = a + b\n' +
      '    - c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 

  opts.operator_position = 'preserve_newline';
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 
  bt('var res = a + b\n' +
      '    - c / d * e % f',
      'var res = a + b - c / d * e % f');  // newline removed 

  opts.preserve_newlines = true;
  opts.operator_position = undefined;
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f');  // unchanged 
 
 // NOTE: Below is different from current behavior
  // Current behavior removes the newline, this is a breaking change, but this seems more consistent and correct.
  // However, if we want to be non-breaking, having the newline be removed here is also acceptable.  
  bt('var res = a + b\n' +
      '    - c / d * e % f',
      'var res = a + b -\n' +
      '    c / d * e % f',);  // newline preserved, operator moved to before the newline

  opts.operator_position = 'before_newline';
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f');  // unchanged 
  bt('var res = a + b\n' +
      '    - c / d * e % f',
      'var res = a + b -\n' +
      '    c / d * e % f',);  // newline preserved, operator moved to before the newline

  opts.operator_position = 'after_newline';
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f',
      'var res = a + b\n' +
      '    - c / d * e % f',);  // newline preserved, operator moved to after the newline
  bt('var res = a + b\n' +
      '    - c / d * e % f');  // unchanged

  opts.operator_position = 'preserve_newline';
  bt('var res = a + b - c / d * e % f');   // unchanged
  bt('var res = a + b -\n' +
      '    c / d * e % f');  // unchanged
  bt('var res = a + b\n' +
      '    - c / d * e % f');  // unchanged
```

What you have forces newlines after all operators even when there were none. What we want is allow newlines in a specific position for operators and optionally move the newline if the code has it in on the wrong side.

olsonpm:
No problem - I added to the confusion by initially coding it that way.  I still have my branch so I can take a look at it this weekend and get back with an up-to-date PR.

bitwiseman:
Let's do the breaking change I showed above.  

It _is_ different, but I'd argue that it is also more correct.  Right now the beautifier punishes people for putting newlines on the wrong side of operators - it looses their intended formatting.  The "break" is that is it will try to preserve their formatting.  I'm okay with that.

olsonpm:
I'm on board, and definitely agree it's more correct.

olsonpm:
Just an fyi - I didn't find time this weekend unfortunately.  I am motivated to get this done though and will give an update by mid next week

bitwiseman:
No problem.  I (obviously) know how it goes.   Thanks!

olsonpm:
@bitwiseman - Took a good stab at this tonight, ran into a case I'm unsure about:

When `operator_position` is set to `remove_newline` What should the following resolve to?
```
var a = b

=== c;
```
The answer depends on whether you want `preserve_newlines` or `operator_position` to take precedence.  [This if statement](https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L389-L394) would need to check for `opt.operator_position !== OPERATOR_POSITION.remove_newline` if we want the newlines to collapse in the above case.

I'll bump this in two weeks if I don't hear back.

bitwiseman:
This is why I didn't include `remove_newline` as an option in my list above.  :smile:  

I figured `preserve_newlines` would take precedence: if set to false existing newlines are ignored, if set to true the behavior can be customized. 

Unless you feel we really need `remove_newline`, I'd drop it.  It seems like something most people would not want - "keep newlines in general, but remove them when they occur before or after an operator"?

olsonpm:
Ah - that's what I get for skimming.  Thanks much for the clarification.

olsonpm:
@bitwiseman - I feel pretty good about this second draft.  Ensuring the operator position logic only applies when preserve_newlines is true simplifies everything, especially tests.

Honestly I was very surprised our breaking change didn't cause a bunch of tests to break.  I was expecting to have to go through and fix a ton of them.

By the way, instead of looking at the diff between the first -> second draft, it may make more sense to diff master -> second draft.

test/data/javascript.js
```diff
@@ -104,6 +107,218 @@ exports.test_data = {
...
+    }, {
+        name: "operator_position option - set to 'before_newline' (default value)",
+        tests: [
+            {
+                comment: 'comprehensive, various newlines',
+                input: inputlib.operator_position.comprehensive,
+                output: [
+                    'var res = a + b - c /',
+                    '    d * e %',
+                    '    f;',
```
bitwiseman: Hmm, the input is: 
```js
var res = a + b
- c /
d  *     e
%
f;
```

For `before_newline`, shouldn't the output be: 
```js
var res = a + b -
    c /
    d * e %
    f;
```

olsonpm: You're right, I'll look into it this weekend.

olsonpm: So I spent some time on this tonight and actually don't agree anymore.  The reason is that if the following holds true:
```
// operator_position = before_newline
a
|| b

// resolves to

a ||
  b
```
Then people will expect the following:
```
// operator_position = after_newline

a ||
b

// resolves to

a
  || b
```

With my understanding of how js-beautify parses text, this last resolution would only be possible by waiting to process the 'b' token.  We'd use the 'wanted_newline' flag to then add a newline before the previous operator token.  This seems messy and unnecessary when preserving the operator_position seems intuitive enough.

Thoughts?

bitwiseman: > Then people will expect the following:

That is exactly what they should expect for `after_newline`.  If they want to preserve the existing newline positioning, they'll use `preserve_newline`.  

I think you're right, it will involve either waiting until the next token to see if it wants a newline and so on.  Messy, but so it goes.   Or you could peek ahead, since all the tokens are already available.

olsonpm: Alright I'll do that - however I disagree with the alternative being synonymous with `preserve_newline` since that will preserve newlines on either side of the operator.

bitwiseman: Peek: yeah, in the javascript module all tokens  are parsed before the formatting handlers run. Here's an example:
https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L749 

This is not the case in css or html modules - I haven't had time to untangle those.

As to behavior: I think I'm just sticking with what I outlined in https://github.com/beautify-web/js-beautify/pull/735#issuecomment-155311837 .   If I'm off target, please call me out.  It's been known to happen. :smile:

olsonpm: haha - I'll keep that in mind.

bitwiseman:
@olsonpm - any update?  I'm driving towards releasing 1.6.0 and I'd love to include this.

bitwiseman:
Okay, I'm pushing this out of 1.6.0.

olsonpm:
I believe the logic you asked for is now implemented.  preserve_newline should be untouched.

test/data/javascript/tests.js
```diff
@@ -316,6 +319,242 @@ exports.test_data = {
...
+                    '    l;'
+                ]
+            }, {
+                comment: 'catch-all, includes brackets and other various code',
+                unchanged: inputlib.operator_position.catch_all
+            }
+        ]
+    }, {
+        name: "reset operator_position",
+        options: [{ name: "operator_position", value: "'before_newline'" }]
```
bitwiseman: Don't need this anymore. The framework now does the resetting for you.

test/data/javascript/tests.js
```diff
@@ -316,6 +319,242 @@ exports.test_data = {
...
+        tests: [
+            {
+                unchanged: inputlib.operator_position.sanity
+            }, {
+                input: inputlib.operator_position.comprehensive,
+                output: inputlib.operator_position.sanity,
+            }
+        ]
+    }, {
+        name: "reset preserve_newlines and operator_position",
```
bitwiseman: Don't need this anymore. The framework now does the resetting for you between each group.

test/data/javascript/tests.js
```diff
@@ -316,6 +319,242 @@ exports.test_data = {
             },
         ],
     }, {
+        name: "operator_position option - ensure no neswlines if preserve_newlines is false",
```
bitwiseman: neswlines

bitwiseman:
This looks great!  Just need the python port and update to cli and we're good to merge.

test/data/javascript/tests.js
olsonpm: Alright fixed the three comments - let me know when you're done and I'll push a fourth draft.  When we're finished I'll rebase them all into a single commit

bitwiseman:
Take a look at 
https://github.com/beautify-web/js-beautify/tree/feature/add-operator-position

olsonpm:
Just dug through the javascript version now and it looks good.  Just to make sure I didn't miss anything, you basically slimmed down logic - turned the cli option to use a dash instead of an underscore, which necessitated the Object.values polyfill instead of keys.  And then you re-organized some of the variables into more appropriate spots.

Thanks for catching the punctuation <--> positional-operators overlap.  Didn't think of that.

bitwiseman:
Re python: Actually. You should be able to read it mostly 1-to-1 from the javascript.

olsonpm:
*Rebasing and cleaning up the commit message now.

If there's anything about my commit message you want changed, it would be easiest if you just create another pull request and modify it yourself to save back-and-forth.  I'm not concerned about having my name on anything.

bitwiseman:
I'm not concerned about the message.

olsonpm:
Alright - I looked at the readme and couldn't figure out why some cli options were left out so I'll leave it be for now.

bitwiseman:
The readme being out of date is totally possible. But unless it applies to this change getting it up to date is not needed.

bitwiseman:
Ah I see I didn't update the help message in python.  Meh, I'll get that in a minute.

olsonpm:
Gotcha - and yeah I just thought updating the readme might make more sense in a separate commit to master.

bitwiseman:
@olsonpm  - thanks again for all your hard work on this. :+1:

NerdPad:
@bitwiseman what is the jsbeautifyrc option for this feature?

olsonpm:
@NerdPad - what version are you using?

NerdPad:
I am using v1.6.2. Here is the output that I am getting:

Source:
```javascript
const result = true
    ? true
    : false;
```

After format:
```javascript
const result = true ? true : false;
```

.jsbeautifyrc

```json
{
    "indent_size": 4,
    "indent_char": " ",
    "eol": "\n",
    "indent_level": 0,
    "indent_with_tabs": false,
    "preserve_newlines": true,
    "max_preserve_newlines": 2,
    "jslint_happy": false,
    "space_after_anon_function": false,
    "brace_style": "collapse-preserve-inline",
    "keep_function_indentation": false,
    "eval_code": false,
    "unescape_strings": false,
    "wrap_line_length": 0,
    "wrap_attributes": "auto",
    "wrap_attributes_indent_size": 4,
    "end_with_newline": true,
    "operator_position": "preserve-newline"
}
```

olsonpm:
Hey thanks much - I'll take a look at this tonight when I'm back from work.

olsonpm:
@NerdPad - 1.6.2 doesn't have this functionality.  Please let me know what led you to believe this was implemented

NerdPad:
@olsonpm I saw the commit was merged, so I assumed it was available in the latest version. What I didn't check was the release date of 1.6.2. Sorry about that.

Do we know when this patch will get in the main release?

olsonpm:
Not a problem - I only asked in case we did anything wrong.

and @bitwiseman will be able to answer that question.

bitwiseman:
Doh, yeah, it went in just after 1.6.2.  I need to get 1.6.3 published. :smile:

NerdPad:
@bitwiseman Please publish 1.6.3. :)

bitwiseman:
Thanks for the reminder. 😄

bitwiseman:
Done! Enjoy!

PR #744, opened 2015-07-22
'<?php ?>' as a single html tag.

yustnip:
Added '?php' to the list of single tags. It will allow to get a pretty code when HTML contains constructions like <?php echo $var; ?>.

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -157,7 +157,7 @@
 
             this.Utils = { //Uilities made available to the various functions
                 whitespace: "\n\r\t ".split(''),
-                single_token: 'br,input,link,meta,source,!doctype,basefont,base,area,hr,wbr,param,img,isindex,embed'.split(','), //all the single tags for HTML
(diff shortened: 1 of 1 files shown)


bitwiseman:
Need some added tests before we can merge this.

Oxicode:
+1

bitwiseman:
@Oxicode - Thanks for the vote!  it looks like @yustnip is not responding, would you like to add tests that show this working?

yustnip:
I don't know how to do it right. :) I can provide screenshots:
https://yadi.sk/i/PMRg8KFLkrsmg
https://yadi.sk/i/GoaJqyBckrsnD

bitwiseman:
Sorry, screenshots don't help at all.  Need to have repeatable tests. 
Make changes to https://github.com/beautify-web/js-beautify/blob/master/test/data/html.js 
Then run `make testj`.  That will generate tests based on the data you added, and then will run them.

bitwiseman:
Looks like this would fix  #826.

PR #751, opened 2015-08-06
Prevent incorrect wrapping of return statements.

glennimoss:
There cannot be a newline between the return keyword and the start of
the return value expression.

### DIFF
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -181,6 +181,8 @@
             Expression: 'Expression' //'(EXPRESSION)'
         };
 
+    var newline_restricted_tokens = /^(continue|break|return|throw)/;
+
     function Beautifier(js_source_text, options) {
         "use strict";
         var output
@@ -415,7 +417,6 @@
             return out;
         }
 
-        var _newline_restricted_tokens = /^(continue|break|return|throw)/;
         function allow_wrap_or_preserved_newline(force_linewrap) {
             force_linewrap = (force_linewrap === undefined) ? false : force_linewrap;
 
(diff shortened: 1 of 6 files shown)


bitwiseman:
This looks like a great change!  
Some minor fixup needed before we can merge it.

js/lib/beautify.js
```diff
@@ -415,6 +415,7 @@
             return out;
         }
 
+        var _newline_restricted_tokens = /^(continue|break|return|throw)/;
```
bitwiseman: This should go with the other tokens and reserved word lists not buried in here.

js/test/beautify-javascript-tests.js
```diff
@@ -1923,6 +1923,7 @@ function run_javascript_tests(test_obj, Urlencoded, js_beautify, html_beautify,
         //.............1234567890123456789012345678901234567890123456789012345678901234567890
         wrap_input_1=('foo.bar().baz().cucumber((fat && "sassy") || (leans\n&& mean));\n' +
                       'Test_very_long_variable_name_this_should_never_wrap\n.but_this_can\n' +
+                      'return between_return_and_expression_should_never_wrap.but_this_can\n' +
```
bitwiseman: Add a test for `throw`.

js/lib/beautify.js
```diff
@@ -426,6 +427,11 @@
             if ((opt.preserve_newlines && current_token.wanted_newline) || force_linewrap) {
                 print_newline(false, true);
             } else if (opt.wrap_line_length) {
+                if (_newline_restricted_tokens.test(flags.last_text)) {
```
bitwiseman: You don't want to use just `last_text` here, you also need `last_type === 'TK_RESERVED'`.

PR #754, opened 2015-08-07
Performance improvement for get_unformatted in html beautify for js

ephox-mogran:
Hi,

We are from Ephox (writers of TinyMCE, EditLive!, Textbox.io etc.) and our users have the ability to embed base 64 images in their content using our editors. These base 64 images can have very, very large attribute values for their src, and we've run into a performance problem with the existing js-beautify codebase. 

Essentially, when it is calling `get_unformatted` currently, it runs through the input until it hits the ending delimiter (e.g. '). Each step of this character iteration, it searches to see if the built up beautified content contains the ending delimiter. This becomes a problem when the attribute values are very large, because each step also involves a search through the entire beautified content to see if the delimiter has been added. This rapidly degrades as the distance to the delimiter increases, and is probably O(n*n).

In our view, you don't need to keep checking the entire beautified content to see if the delimiter is there, only the last m characters (where m is the delimiter length). Therefore, we suggest introducing a second string which has a maximum length of the delimiter's length. If you add a character to it, then a character is dropped from its start. Using this approach, we can greatly decrease the amount of time it takes to check if we have reached the ending delimiter without breaking the algorithm.

Of course, this approach only works if the number of characters added at each step is less than the delimiter's length. Aside from the `{{` and `}}` pairing where delimiters are specifically ignored, the only additions each step are whitespace (which shouldn't really be compared with the delimiter anyway) and the single input character at the current position. Therefore, it would make sense that the algorithm's correctness should be preserved, but if we have missed something, let us know.

All of the tests appear to still pass, and we have introduced a new performance test for html beautify which demonstrates what the problem with the current implementation is.

We would appreciate it if you would consider merging this into your main codebase. It should improve the performance for all users.

Thanks,
Morgan.
Ephox engineer.

Note, there is a `substr` call which is given a negative number. This works on all browsers except IE <= 8. If this is a problem for your browser support, we can change it.


bitwiseman:
Yes, we maintain backward support to ancient IE6.  I don't love it either.

js/test/node-beautify-perf-tests.js
```diff
@@ -20,7 +20,7 @@ function node_beautifier_tests() {
     //warm-up
     js_beautify(data, options);
     js_beautify(data_min, options);
-
+
```
bitwiseman: You can clean this up.

bitwiseman:
This looks like a good change.  Thanks for taking the trouble to contribute to this project, including expanded tests. 

Please make the minor tweaks needed and I'll merge.

js/lib/beautify-html.js
```diff
@@ -569,15 +569,34 @@
...
+            this.get_unformatted = function(delimiter, orig_tag) { //function to return unformatted content in its entirety
                 if (orig_tag && orig_tag.toLowerCase().indexOf(delimiter) !== -1) {
                     return '';
                 }
                 var input_char = '';
                 var content = '';
-                var min_index = 0;
                 var space = true;
+
+                var matcher = tokenMatcher(delimiter);
```
bitwiseman: Change this var name to `delimiterMatcher`.

ephox-mogran:
Actually, the substr replacement for older browsers isn't quite right. If delimiter is longer than newToken, then we get the wrong result. I'll update the PR tomorrow with the fix.

ephox-mogran:
*Should* be right now.

bitwiseman:
Thanks for the great contribution! 
Before you change:
```
html-beautify (base64 image) x 0.20 ops/sec ±0.55% (5 runs sampled)
```
After your change:
```
html-beautify (base64 image) x 18.86 ops/sec ±2.98% (36 runs sampled)
```
:trophy:

Issue #618, opened 2015-02-06
Extra space after pseudo-elements within :not()


jessefisher:
Beautifying this code:
`&:not(:last-child) {}`
Results in this:
`&:not(: last-child) {}`

The extra space causes an error, which tries to identify the line above as an invalid property error.

Thanks for the great extension!

bitwiseman:
Is this the same as #557?

jessefisher:
If the patch for [#557](https://github.com/beautify-web/js-beautify/issues/557) fixes the space between all colon and pseudo-elements then yes. In this instance the extra space has only occured on `:last-child` when inside `:not()`.

bitwiseman:
Um... is there a patch for #557 that I don't know about?

jessefisher:
No, sorry I meant you might be able to fix it with the same solution.

bitwiseman:
Okay, good to have another example.

bitwiseman:
This appears to be fixed.  On http://jsbeautifier.org/ this remains unchanged:
```html
<style>
    &:not(:last-child) {}
</style>
```
Can anyone verify?

bitwiseman:
@jessefisher, This appears to be fixed.  On http://jsbeautifier.org/ this remains unchanged:
```html
<style>
    &:not(:last-child) {}
</style>
```
Can you verify?

corgrath:
This is still broken.

This code:

```
  <style>
      div {
          &:not(:first-of-type) {
              background: red;
          }
      }
  </style>
```

Results in:

```
  <style>
      div {
          &:not(: first-of-type) {
              background: red;
          }
      }
  </style>
```

brianfryer:
I am also experiencing this issue.

maelfrize:
Same here

alan-cole:
Same.

I find if compiling the following:

```
[type="checkbox"]:not(:checked) {
  background-color: transparent;
}

@mixin hello {
  [type="checkbox"]:not(:checked) {
    background-color: transparent;
  }
}
```

The output is:

```
[type="checkbox"]:not(:checked) {
  background-color: transparent;
}

@mixin hello {
  [type="checkbox"]:not(: checked) {
    background-color: transparent;
  }
}
```

Notice the first `:not(:checked)` outside the braces works, the second one inside the @mixin{} braces doesn't work.

hmaesta:
[...]

PR #760, opened 2015-08-20
Bugfix for #618

corgrath:
Hey. This is a pull request to solve issue https://github.com/beautify-web/js-beautify/issues/618

I know there is a setup you need to do, that includes Python, but since I have no Python experience at all, I had problems setting the test environment up. But I saw that Travis-ci, so hopefully that can help with testing. If not, is there a chance any main developers can help verify that this fix doesn't break anything?

### DIFF
--- a/js/lib/beautify-css.js
+++ b/js/lib/beautify-css.js
@@ -364,7 +364,7 @@
             } else if (ch === ":") {
                 eatWhitespace();
                 if ((insideRule || enteringConditionalGroup) &&
-                    !(lookBack("&") || foundNestedPseudoClass())) {
+                    !(lookBack("&") || lookBack("(") || foundNestedPseudoClass())) {
                     // 'property: value' delimiter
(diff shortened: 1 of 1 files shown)


corgrath:
The problem is that I haven't added any new test cases to ensure this change will preserve. Is this something you can help me with?

This could be a test case:
```
div 
{
	&:not(:first-of-type   ) { background: red;	}
}
```
Should become
```
div {
	&:not(:first-of-type) {
		background: red;
	}
}
```

bitwiseman:
Please look at https://github.com/beautify-web/js-beautify/blob/master/test/data/css.js. Add you test using that json, then run make. This will generate tests for python and javascript.

Thanks for taking the time to contribute!

corgrath:
I am closing this.

I cannot build the project, as I am getting problems running make in windows.

Perhaps someone else in the future has the time to create the necessary tests.

Issue #763, opened 2015-08-24
Missing 'var' in beautify-html.js; breaks strict mode


pekim:
https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify-html.js#L527

```input_char = this.input.charAt(this.pos);```
should be
```var input_char = this.input.charAt(this.pos);```

This breaks when used in a node app, and node is run with `--use-strict` flag.

PR #764, opened 2015-08-24
Add missing 'var' declaration.

pekim:
Fixes issue #763.

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -524,7 +524,7 @@
                     matched = false;
 
                 this.pos = start_pos;
-                input_char = this.input.charAt(this.pos);
+                var input_char = this.input.charAt(this.pos);
                 this.pos++;
 
                 while (this.pos <= this.input.length) {


js/lib/beautify-html.js
```diff
@@ -524,7 +524,7 @@
                     matched = false;
 
                 this.pos = start_pos;
-                input_char = this.input.charAt(this.pos);
+                var input_char = this.input.charAt(this.pos);
```
corgrath: A question. JSLint practices usually says that declarations should be in the top of function blocks due to hoisting. It seems like this practice is upheld in this function block as well.

Should the declaration of 'input_char' be moved to the rest of the declaration chain?

pekim: If the file followed that convention it would make sense to follow it. However neither this file nor the project as a whole appear follow the convention.

timhudson:
Bump, I'd like to see this get merged

bitwiseman:
Yeah, this project is weak on code style and lint checks.
I'd be happy to have someone enact them.

PR #773, opened 2015-09-15
added no spaces before conditional; added no newline for empty_braces

melfid:
Added option to python to allow space before conditional.
Added new feature to keep braces on same line when braces are empty; Many times in JavaScript empty objects are passed as function variables, splitting onto newlines is visually disruptive.


bitwiseman:
Thanks for contributing!

Please add tests and javascript implementation.

bitwiseman:
Closing for lack of activity.  Will reopen upon request.

PR #777, opened 2015-09-17
is_whitespace for loop incrementing wrong variable

ephox-luke:
Without this fix `is_whitespace` only ever checks the first character. Called when the first token is text content; If that token has leading whitespace the whole token is ignored (content loss). Bug not present in python implementation.


bitwiseman:
Thanks!

PR #785, opened 2015-10-12
Resolve #784 Handlebars {{else}} tag not given a newline

### LINKED ISSUE #784: Handlebars {{else}} tag not given a newline
Take following input. Handlebars if-else, with `div` tags within each condition branch:
```
{{#if condition}}<div></div>{{else}}<div></div>{{/if}}
```

This will generate the following output:
```
{{#if condition}}
    <div></div>{{else}}
    <div></div>
{{/if}}
```

I would have expected this.. `{{else}}` tag drops to same indentation level as the `{{#if}}`:

```
{{#if condition}}
    <div></div>
{{else}}
    <div></div>
{{/if}}
```

tmarchant:
Solves the problem shown in the issue.

A newline will only be placed before an `{{else}}` tag when it's corresponding `{{#f}}` tag is on a separate line.

My method for finding whether or not an `{{#if}}` tag is on the same line is a bit lame/naive (checking back on previously-output tokens) and could be improved by someone who knows the code base better

N.B only fixed JS version. Python not touched

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -831,6 +831,21 @@
                     multi_parser.current_mode = 'CONTENT';
                     break;
                 case 'TK_TAG_HANDLEBARS_ELSE':
+                    // Don't add a newline if opening {{#if}} tag is on the current line
+                    var foundIfOnCurrentLine = false;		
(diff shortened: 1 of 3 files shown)


bitwiseman:
There's no python html, so you're good.  Thanks!

PR #790, opened 2015-10-19
Add javascript2img unpacker

sfan5:
Not sure if this unpacker is worth adding as the packer is probably not widely used.
Even though the cli version doesn't seem to use the unpackers the code works with nodejs too.
When running under nodejs it requires *pngjs2* but it fails gracefully and warns the user if that module is not installed.

(IMO It's worth adding solely because [their website](http://javascript2img.com/) says that <code>JsBeautifier can't deobfuscate our code!</code> :smirk:)


bitwiseman:
Hilarious.  :smile:  
There have a been a whole pile of "unpacker" requests recently.  It has me thinking that they should move to a separate project, since there are many many ways to pack javascript.

Issue #803, opened 2015-11-12
Misformating "0b" Binary Strings


mpelikan:
ES6 allows binary strings (numeric literal form) to be formatted in "0b1010".

bitwiseman:
Fixed by 80c8368

PR #804, opened 2015-11-12
Fixing #803

mpelikan:
Added support for binary numeric literal form.

### DIFF
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -1682,12 +1682,10 @@
                     parser_pos += 1;
                     if ( /[b]/.test(input.charAt(parser_pos)) ) {
                         local_digit = digit_bin;
-                    }
-                    else {
+                    } else {
                         if ( /[o]/.test(input.charAt(parser_pos)) ) {
                             local_digit = digit_oct;
-                        }
-                        else {
+                        } else {
                             local_digit = digit_hex;
                         }
                     }
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -1532,6 +1532,7 @@
 
         var whitespace = "\n\r\t ".split('');
         var digit = /[0-9]/;
+        var digit_bin = /[01]/;
         var digit_oct = /[01234567]/;
         var digit_hex = /[0123456789abcdefABCDEF]/;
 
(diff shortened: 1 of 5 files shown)


js/lib/beautify.js
```diff
@@ -1673,13 +1674,23 @@
...
+                    // switch to hex/oct/bin number, no decimal or e, just hex/oct/bin digits
                     allow_decimal = false;
                     allow_e = false;
                     c += input.charAt(parser_pos);
                     parser_pos += 1;
-                    local_digit = /[o]/.test(input.charAt(parser_pos)) ? digit_oct : digit_hex;
+                    if ( /[b]/.test(input.charAt(parser_pos)) ) {
+                        local_digit = digit_bin;
+                    }
+                    else {
```
bitwiseman: Minor formatting and style changes:
```js
} else if ( /[o]/.test(input.charAt(parser_pos)) ) {
    local_digit = digit_oct;
} else {
    local_digit = digit_hex;
}
```

python/jsbeautifier/__init__.py
```diff
@@ -1498,16 +1499,19 @@ def __tokenize_next(self):
...
                 allow_e = False
                 c += self.input[self.parser_pos]
                 self.parser_pos += 1
-                if re.match('[o]', self.input[self.parser_pos]):
-                    local_digit = self.digit_oct
+                if re.match('[b]', self.input[self.parser_pos]):
+                    local_digit = self.digit_bin
                 else:
-                    local_digit = self.digit_hex
+                    if re.match('[o]', self.input[self.parser_pos]):
```
bitwiseman: Same thing here as in js,  use `elif`.

bitwiseman:
Style changes and tests needed. 
Thanks!

mpelikan:
Tests already provided.

bitwiseman:
For `0b` case?  I don't see them.

mpelikan:
Oops. Sorry. I changed generated tests, not the source. Added test and fixed style to comply.

js/lib/beautify.js
```diff
@@ -1673,13 +1674,21 @@
...
+                if (c === '0' && parser_pos < input_length && /[Xxob]/.test(input.charAt(parser_pos))) {
+                    // switch to hex/oct/bin number, no decimal or e, just hex/oct/bin digits
                     allow_decimal = false;
                     allow_e = false;
                     c += input.charAt(parser_pos);
                     parser_pos += 1;
-                    local_digit = /[o]/.test(input.charAt(parser_pos)) ? digit_oct : digit_hex;
+                    if ( /[b]/.test(input.charAt(parser_pos)) ) {
+                        local_digit = digit_bin;
+                    } else {
```
bitwiseman: Use `} else if {`, then this will be good to merge.

mpelikan: I disagree. That is not officially valid Javascript. There is no "else if" in the language. The "else" is always put in a else block. While the language will handle "else if", I am being explicit for clarity sake and to allow to extension of the else block.

This seems to be a nitpick for accepting a pull request.

bitwiseman: This project uses `} else if (...) {` where reasonable and appropriate, this is one of those cases. 

We attempt to maintain parity between python and javascript versions of the code.  Not always possible, but where you see `elif ...:` in python, you should see `} else if (...) {` in javascript.

bitwiseman:
Made the adjustment and committed.  Thanks for contributing!

Nogrod:
breaks hex formatting -> example: 0xb401c4 will become 0x b401c4

Issue #810, opened 2015-11-18
ES6 Object Shortand Indenting Weirdly Sometimes


timothyeburke:
```
// original
function badFormatting() {
    return {
        a,
        b: c,
        d: e,
        f: g,
        h,
        i,
        j:k
    }
}

function goodFormatting() {
    return {
        a: b,
        c,
        d: e,
        f: g,
        h,
        i,
        j:k
    }
}


// output
function badFormatting() {
    return {
        a,
        b: c,
            d: e,
            f: g,
            h,
            i,
            j: k
    }
}

function goodFormatting() {
    return {
        a: b,
        c,
        d: e,
        f: g,
        h,
        i,
        j: k
    }
}
```

In ES6, the object shorthand is getting weirdly indented when preserving new line, but only in the case that the object's first token is the shorthand syntax. This syntax is documented here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

If you feel this is a legit bug, I'll be glad to look into fixing it.

bitwiseman:
It is a legit bug.
Probably related to https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L741 .
It's the `:` in the first field that we look for. Before ES6 that was a reasonable assumption.

PR #812, opened 2015-11-18
Resolving #810: ES6 Object Literal Shorthand Detection

timothyeburke:
I had to update some of the decode tests - they no longer made sense once wrapped in the second part of the test and were failing. Let me know if this works for you.

### DIFF
--- a/js/test/beautify-javascript-tests.js
+++ b/js/test/beautify-javascript-tests.js
@@ -1277,6 +1277,32 @@ function run_javascript_tests(test_obj, Urlencoded, js_beautify, html_beautify, 
             'a.b("c",\n' +
             '    () => d.e\n' +
             ')');
+        
+        // Issue 810 - es6 object literal detection
+        bt(
+            'function badFormatting() {\n' +
+            '    return {\n' +
+            '        a,\n' +
+            '        b: c,\n' +
+            '        d: e,\n' +
+            '        f: g,\n' +
+            '        h,\n' +
+            '        i,\n' +
+            '        j: k\n' +
+            '    }\n' +
+            '}\n' +
+            '\n' +
+            'function goodFormatting() {\n' +
+            '    return {\n' +
+            '        a: b,\n' +
+            '        c,\n' +
+            '        d: e,\n' +
+            '        f: g,\n' +
+            '        h,\n' +
+            '        i,\n' +
+            '        j: k\n' +
+            '    }\n' +
+            '}');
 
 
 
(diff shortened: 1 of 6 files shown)


bitwiseman:
I'm not opposed to test changes in general, but I don't understand why the change to the unencode tests is needed.  I don't think that code path is touched by your change. 

Other than that, it looks reasonable to me. There's also the `myProperty() { ... },` syntax to contend with, but if you want to open a separate issue and solve that later, that'd be fine.

js/test/beautify-javascript-tests.js
```diff
@@ -1277,6 +1277,15 @@ function run_javascript_tests(test_obj, Urlencoded, js_beautify, html_beautify,
             'a.b("c",\n' +
             '    () => d.e\n' +
             ')');
+        
+        // Issue 810 - es6 object literal detection
+        bt(
+            '{\n' +
+            '    a,\n' +
+            '    b: c,\n' +
+            '    d: e,\n' +
```
bitwiseman: You might as well put both inputs that you listed in the issue report in here.  They were just a little more interesting than this.

test/data/javascript.js
```diff
@@ -1196,6 +1196,17 @@ exports.test_data = {
                     ')'
                 ]
             },
+            {
+                comment: "Issue 810 - es6 object literal detection",
+                unchanged: [
```
bitwiseman: You might as well put both inputs that you listed in the issue report in here. They were just a little more interesting than this - worth covering.

timothyeburke: Will do that.

timothyeburke:
As for the updated test - it's because that test is currently in the format:
```
"foo", 'bar', "baz"
```
and on https://github.com/beautify-web/js-beautify/blob/master/test/template/python-javascript.mustache#L1237 that input gets wrapped in a `{ }` block, meaning it is now
```
{ "foo", 'bar', "baz" }
```
and then beautified and treated as an object literal. Replacing those commas with semicolons changes the test to treat them as single line expressions rather than part of an object literal.

bitwiseman:
Ah, okay.  The javascript version of the test passed? 
If so, then change the content back for both test and in the python test mustache change `bt(...)` to `test_fragment(...)`.  That will avoid the `{ }` wrapping.  

Also, as a last point, an you take your changes and rebase them into one commit?

timothyeburke:
Done and done! My team currently uses this (as part of `grunt-jsbeautifier`) for our es5 files, but I want to get through the issues that are keeping us from using it for our es6. Glad I can help, and thanks for helping me through this one.

bitwiseman:
@timothyeburke - High quality and low friction.  Glad to have you do anymore fixes you choose.  :smile:

PR #820, opened 2015-11-30
package.json: Remove "preferGlobal" option

feross:
We use `js-beautify` internally, as part of [`standard`](http://standardjs.com/). Our users get this warning whenever they `npm install standard`:

```
npm WARN prefer global js-beautify@1.5.10 should be installed with -g
```

It's no longer a best-practice to use `preferGlobal` for npm packages that merely contain a command line program. The command line program should be *the only* way to use the package.

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "1.5.10",
   "description": "jsbeautifier.org for node",
   "main": "js/index.js",
-  "preferGlobal": true,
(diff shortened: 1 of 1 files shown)


bitwiseman:
Seems reasonable.  it is up to the user if they want to install global or not.

feross:
Cool, thanks!

PR #831, opened 2015-12-13
Improve Python packaging

smurfix:
The first patch adds Debian packaging support.
The second patch fixes setup.py by using setuptools instead of distutils, and adding in py.test so that "setup.py test" works with Python 3.4+.


bitwiseman:
To be honest, I don't know enough about this to know if it is correct or not.  Could you provide some pointers to documentation I can read to understand this domain?

smurfix:
Hello Liam,
>
> To be honest, I don't know enough about this to know if it is correct
> or not. Could you provide some pointers to documentation I can read to
> understand this domain?
>
Other than "it works"? ;-)

setuptools vs. distutils:
https://python-packaging-user-guide.readthedocs.org/en/latest/current/
"""
Packaging Tool Recommendations
* Use setuptools to define projects and create Source Distributions.
"""

Debian packaging: the patch includes a minimal set of packaging files to
allow somebody to cleanly install the tools locally.
I'm a long-time Debian Developer, so I'm supposed to know what I'm
doing. :-P
References (a long and probably-boring read):
https://wiki.debian.org/Python/Packaging
https://www.debian.org/doc/packaging-manuals/python-policy/index.html
https://www.debian.org/doc/debian-policy/

-- 
-- Matthias Urlichs

bitwiseman:
Yes, other than it works. :) 

Okay, I'm getting a sense for the change now.

Do we need to update https://github.com/beautify-web/js-beautify/blob/master/CONTRIBUTING.md#publishing-a-release to include how to publish the Debian package? 

You've updated this to install the python version of the tool, but the a javascript/nodejs version is arguably better and includes the html beautifier.  Or should the Debian package install both?

smurfix:
On 14.12.2015 23:33, Liam Newman wrote:
>
> Do we need to update
> https://github.com/beautify-web/js-beautify/blob/master/CONTRIBUTING.md#publishing-a-release
> to include how to publish the Debian package?
>
Not really. It's just minimal packaging at this point. Getting it into
Debian proper would entail a bit more work.

Updating the Debian package's version would be a nice touch (it's in the
first line of …/debian/changelog), but again, doing this "properly"
requires running the "debchange" tool. I assume you don't want to force
people to do the release on a Debian machine. ;-)
>
> You've updated this to install the python version of the tool, but the
> a javascript/nodejs version is arguably better and includes the html
> beautifier. Or should the Debian package install both?
>
That should be in a different package, as they have different
dependencies (the Python version requires … well, Python, while the
nodejs version needs node.js obviously).

I'm not at all familiar with how to package Node for Debian, though, and
most people don't bother and install via node.js instead of through the
Debian packaging system.

-- 
-- Matthias Urlichs

bitwiseman:
Could you add instructions to https://github.com/beautify-web/js-beautify/blob/master/CONTRIBUTING.md#publishing-a-release on how someone would update this?  Whatever you think is reasonable.   Including the note that Debian packages may lag behind regular release.  Then I'll go ahead and merge.  

Thanks for your patience and contribution.

bitwiseman:
@smurfix - I guess I'll just to take it as a given that you'll maintain the debian packages.

PR #853, opened 2016-01-26
Destructuring, importing, and short objects

### LINKED ISSUE #114: Format json in line
At present, jsbeautify will format json into multi-lines. For example:

    var items = nativeUsers.find({ email: 'aaa@aaa.com' }, _).toArray(_);

will be formatted as:

    var items = nativeUsers.find({
        email: 'aaa@aaa.com'
    }, _).toArray(_);

But I think it's kept inline will be better for this code.

Will jsbeautify handle it like this:

1. if `{` is at the line end, it will be formatted into multi-lines:

    var items = nativeUsers.find({
                       email: 'aaa@aaa.com' }, _).toArray(_);

2. if not, format them inline

So, we can control if the json be formatted inline or multiline

### LINKED ISSUE #315: Option to preserve or inline "short objects" on a single line
Typically the JS Beautifier expands all JS object declarations into multiple lines, each line having a key value pair. However, this is overly verbose for many situations (e.g. short objects which could easily fit on a single line).

Additional details found here: https://github.com/einars/js-beautify/pull/55

### LINKED ISSUE #370: Ignore object formatting option?
If I have an array of objects that looks like this: 

```js
var methods = [
  { name: "to" },
  { name: "step" },
  { name: "move" },
  { name: "min" },
  { name: "max" },
  { name: "center" },
  { name: "sweep" },
  { name: "stop" },
  { name: "clockWise" },
  { name: "cw" },
  { name: "counterClockwise" },
  { name: "ccw" },
  { name: "write" }
];
```

It will be transformed into this: 

```js
var methods = [{
  name: "to"
}, {
  name: "step"
}, {
  name: "move"
}, {
  name: "min"
}, {
  name: "max"
}, {
  name: "center"
}, {
  name: "sweep"
}, {
  name: "stop"
}, {
  name: "clockWise"
}, {
  name: "cw"
}, {
  name: "counterClockwise"
}, {
  name: "ccw"
}, {
  name: "write"
}];
```

Which is almost twice the number of lines and arguably less readable. Is there an existing solution to this problem?

### LINKED ISSUE #382: Newline inserted after ES6 module import/export
Hey,

I'm running jsbeautifier on a collection of Ember.JS ES6 modules and I've noticed a small issue with export statements.

Suppose I have a module that exports as below

```javascript
export default DS.FixtureAdapter.extend();
```

jsbeautifier will add a newline after export

```javascript
export
default DS.FixtureAdapter.extend();
```

It's a minor problem that is only an issue for us as we enforce a jsbeautifier run before a commit is accepted. So, if a developer removes the newline, the file in question will be included in the commit even though it may have nothing to do with the changes being commited.

bitwiseman:
Fixes #114 
Fixes #315 
Fixes #370
Fixes #382

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # JS Beautifier
 [![Build Status](https://img.shields.io/travis/beautify-web/js-beautify/master.svg)](http://travis-ci.org/beautify-web/js-beautify)
+[![Build status](https://ci.appveyor.com/api/projects/status/5bxmpvew5n3e58te/branch/master?svg=true)](https://ci.appveyor.com/project/beautify-web/js-beautify/branch/master)
 [![NPM version](https://img.shields.io/npm/v/js-beautify.svg)](https://www.npmjs.com/package/js-beautify)
 [![Download stats](https://img.shields.io/npm/dm/js-beautify.svg)](https://www.npmjs.com/package/js-beautify)
 [![Join the chat at https://gitter.im/beautify-web/js-beautify](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/beautify-web/js-beautify?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
--- a/index.html
+++ b/index.html
@@ -356,6 +356,7 @@
 
                 <select id="brace-style">
                     <option value="collapse">Braces with control statement</option>
+                    <option value="collapse-preserve-inline">Braces with control statement, preserve inline</option>
                     <option value="expand">Braces on own line</option>
                     <option value="end-expand">End braces on own line</option>
                     <option value="none">Attempt to keep braces where
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -61,7 +61,7 @@
     space_after_anon_function (default false) - should the space before an anonymous function's parens be added, "function()" vs "function ()",
           NOTE: This option is overriden by jslint_happy (i.e. if jslint_happy is true, space_after_anon_function is true by design)
 
-    brace_style (default "collapse") - "collapse" | "expand" | "end-expand" | "none"
+    brace_style (default "collapse") - "collapse-preserve-inline" | "collapse" | "expand" | "end-expand" | "none"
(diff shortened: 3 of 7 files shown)


ksloan:
Awesome! All praise @bitwiseman! Thanks so much for doing this, works wonderfully.

fyockm:
Oh happy day! Can't wait to try this out!

underscorebrody:
@bitwiseman Thanks for doing this! It helps a lot!

One thing I'm still seeing is that this:

```
function something({a, b, c}, d){
  //whatever
}
```

is becoming this:

```
function loadWasteSheetData({
  a, b, c
}, d) {
  //whatever
}
```

This is using ES6 object destructuring in the function arguments, so I thought it might apply here. But let me know if this is a separate issue entirely. Thanks!

bitwiseman:
@underscorebrody - No, I was expecting it to fix the input you provided.  It looks like it comes out right locally.  Did you turn on the new setting?

underscorebrody:
Okay this was my fault, I ran the beautify task against some code that had been beautified already on the old version so it didn't get handled the right way. Sorry!

bitwiseman:
@underscorebrody - No problem, thanks for jumping in and testing right away.  Please report any other issues you find. :smile:

underscorebrody:
Discovered https://github.com/beautify-web/js-beautify/issues/854 while testing as well, but i don't think it's directly related to this change

jwhitmarsh:
@bitwiseman I've just read through #114,  #315, #370 and #382, and _really_ appreciate the work you've put into supporting this! 

However, with v1.6.2 

``` 
import {join, other} from 'path'

// is becoming

import {
  join, other
}
from 'path';
```

Is that expected, or an issue with my setup?

bitwiseman:
Probably a config issue: you need to set `brace_style` to `collapse-preserve-inline` to get the inline behavior.

https://github.com/beautify-web/js-beautify/blob/master/test/data/javascript/tests.js#L1572

monkindey:
Thanks @bitwiseman Well done;
but is it when i type the object like 
```javascript
 //input
var test = {a: 1, b:2}
```

it did't transform to 

```javascript
//output
var test = {
   a: 1,
   b: 2
}
```

but i expect output like above

> thanks your project

PR #861, opened 2016-01-31
Add missing 'collapse-preserve-inline' option to js module

vekat:
I wasn't able to use the new `collapse-preserve-inline` using the npm module, so I decided to take a look and found out it was missing.


bitwiseman:
Ah, I missed that this option was sanitized here. Thanks!

PR #867, opened 2016-02-02
1.6.2 in readme

jayceekay:
@bitwiseman


bitwiseman:
There's been no significant update to the readme since 1.6.0.   Are you saying the readme should always say the current version?

jayceekay:
i only noticed because i updated and did a sanity check to be sure it pulled down the latest, noticed it said 1.6.0 and not 1.6.2. no strong opinion on the matter, just thought that's what it was there for :)

bitwiseman:
Fair point. #868

Issue #844, opened 2016-01-19
Bug in Less mixins


baivong:
I'm using package HTML-CSS-JS Prettify in Sublime Text 3.

Before:
```less
.mixin(@abc, @xyz) {
    @{abc}: @xyz;
}
```

After:
```less
.mixin(@abc, @xyz) {
    @ {
        abc
    }
    : @xyz;
}
```

bitwiseman:
Cool thanks.

agulado:
+1
less

Before：
```css
    .loop_section_top_ul_li(@i) when (@i>0) {
    	section.top ul li.l@{i} a{
        	background-position: ((@i)-1)*-100px -20px;
    	}
    	section.top ul li.l@{i} a.now{
        	background-position: ((@i)-1)*-100px 0;
    	}
    	.loop_section_top_ul_li((@i)-1);
    }
```

After：
```css
    .loop_section_top_ul_li(@i) when (@i>0) {
        section.top ul li.l@ {
            i
        }
        a {
            background-position: ((@i)-1)*-100px -20px;
        }
        section.top ul li.l@ {
            i
        }
        a.now {
            background-position: ((@i)-1)*-100px 0;
        }
        .loop_section_top_ul_li((@i)-1);
    }
```

agulado:
3ks

Issue #628, opened 2015-03-03
breaks less css variable interpolation


thisguychris:
Using variable interpolators for less breaks during compile.

this:
```
// Variables
@mySelector: banner;

// Usage
.@{mySelector} {
  font-weight: bold;
  line-height: 40px;
  margin: 0 auto;
}
```

becomes:

```
// Variables
@mySelector: banner;
// Usage
.@ {
  mySelector
}
{
  font-weight: bold;
  line-height: 40px;
  margin: 0 auto;
}
```

expected:

```
.@{mySelector} {
```

should be on the same line.

PR #873, opened 2016-02-12
Fix Issue #844

orottier:
Less beautifier: 

> Properties can be interpolated, e.g. @{prefix}-property: value


orottier:
Iterative development, we're getting there

bitwiseman:
Okay. :+1:

orottier:
Probably related: #628

bitwiseman:
@orottier - Fantastic start!

It looks like you have a solution that stops the breaking result.  But it doesn't look like this beautify unformatted input.  For example, what happens when your input is `tag {@{prop}: none;}`?  
Could you add some additional tests for a wider set of inputs?

bitwiseman:
Great, thanks!

PR #878, opened 2016-02-19
Fix incomplete list of non-positionable operators (python lib)

orottier:
Now it is the same as the JS list:
js/lib/beautify.js:1746

This caused operations like 'a -= 2;' to be changed to
'a - = 2;', which breaks


bitwiseman:
@orottier,
Great! Please add a test that uses these symbols to prevent regressions.

Issue #916, opened 2016-04-08
Option to put HTML attributes on their own lines, aligned


JVimes:
Suggest an option to put HTML attributes on their own lines, aligned with each other. For example, turn this:
``` html
<button id="start-signout" class="button button-full" ng-click="vm.signout()" click-animation translate>AUTH.SIGN_OUT</button>
```
into this:
``` html
<button id="start-signout"
        class="button button-full"
        ng-click="vm.signout()"
        click-animation
        translate>AUTH.SIGN_OUT</button>
```

This [has been requested](https://github.com/Microsoft/vscode/issues/2204) and up-voted in a dependent product, Visual Studio Code.

siulung4073:
+1

bitwiseman:
Fixed by @Lukinos.  Thanks!

hneuer:
Great work!

Is there an ETA for the 1.6.5 release? I would really like to see the force-align feature get into vscode!

zourun:
vscode 1.23.1  
"html.format.wrapAttributes": "force-aligned",
not works

Seebiscuit:
VSCode 1.23.1
"html.format.wrapAttributes": "force-aligned" 
Not working in Vue SFC templates

Cristy94:
It would be nice if we could bind this "html.format.wrapAttributes" command to a keyboard shorcut.

PR #936, opened 2016-05-12
Adds #916

cfischer:
wrap_attributes: 'force',
wrap_attributes_indent_size: 'align',
reformat_unformatted_tags: true


cfischer:
Closes #916

bitwiseman:
Hello, I'm sorry for the delay.  I've had no time to review changes or to fix the issue with the CI tools. Please be patient.

bitwiseman:
It looks like we still have failures.  Please take a look when you get a chance.

bitwiseman:
@cfischer - please take a look and add tests.  You can run `./build` to verify your feature is working at expected.

Lukinos:
@bitwiseman - I did not see this pull request, and hence I also coded this feature (https://github.com/Lukinos/js-beautify/commit/341052954730390714e32846419097889f807d58). But what I wanted to know is how it should be tested. Because, the test ecosystem seems to be too rigid.

Concretely, I wanted to add my tests to the 'wrap_attributes' tests (name: AttributeWrap), however the 'indent'_attr' value is fixed through all tests (for one entry matrix) and as we want to test something where the 'indent_attr' is dynamic for each test, the tests break.

What would be the right way to do ? Creating several specific groups just for this new attribute ? Something else ?

bitwiseman:
@Lukinos - Please create a new pull request and try adding one test where you think it should go (even if it fails). Then I'll be able to give you concrete feedback.

Issue #917, opened 2016-04-11
Possible to add an indent_inner_inner_html option? (Prevent indenting second-level tags)


BevansDesign:
I'm new to Github, and don't really have any JS experience, but I was hoping that this could be added as an option:

Currently, with indent_inner_html, we can choose if we want to indent the first-level tags within the <HTML> tags. However, it'd be great if an "indent_inner_**inner**_html" option existed, to prevent the indentation of second-level tags within <HEAD> and <BODY>.

For example, instead of this:
```
<body>
    <div>
        <div>
        </div>
    </div>
</body>
```

I'd like to have this:
```
<body>
<div>
    <div>
    </div>
</div>
</body>
```

Thanks for your consideration! This extension is great.

PR #946, opened 2016-06-02
Add options to prevent indent of body and head contents

spontaliku-softaria:
Adds #917. Options indent_body_inner_html and indent_head_inner_html.


bitwiseman:
Hey, great start.  But think one setting would be enough.  The original request was for "indent_inner_inner_html".  I think the original name is a but odd but still better than adding two settings.

Also, please add tests for the new settings.  Then we can merge this.  Thanks!

spontaliku-softaria:
Actually for my project I need to manage both head and body, so I decided implement both. Is it OK? One can achieve #917 using just "indent_body_inner_html".

I've added some tests.

bitwiseman:
Two settings sounds good then.  And the tests look like good.  Thanks!

PR #949, opened 2016-06-07
Fix errors being thrown on non-Latin tag characters

TimothyGu:
Example:

```html
<中文></中文>
```

Yes, this is not valid HTML according to the spec, but this commit makes the beautifier more robust against possible errors.


bitwiseman:
@TimothyGu please run "./build" and then amend your commit.

TimothyGu:
Sorry about that. Fixed.

Issue #551, opened 2014-10-03
Integrating editor config


ilanbiala:
What do you guys think about taking into account editorconfig files when beautifying. I'm not sure how many people use this in a package/plugin/build proccess way, but when JSBeautify is used to beautify a file, some settings that editorconfig sets up are disregarded and overriden like trailing newlines. Do you guys think it would be a good idea to allow people to pass in a .editorconfig file programmatically and for the CLI look for one in the cwd and take these settings into account to avoid that code being duplicated everywhere else?

dcharbonnier:
editorconfig support would be very very nice

bitwiseman:
Please help!  Submit a pull request!

justrossthings:
this is would be really useful to contribute to open source projects. Make it pls! ლ ( ◕  ᗜ  ◕ ) ლ

mitermayer:
Has anyone create a PR for this? I am happy to create one if no-body has done it yet.

ethanluoyc:
I am starting to implement the editorconfig functionality. Anybody who wants to review my code?

ethanluoyc:
I have implemented the feature. Please feel free to try it out and comment on it.

PR #969, opened 2016-07-11
Add support for EditorConfig

ethanluoyc:
This is the very first commit to support #551. Now you can pass in an `.editorconfig` file(s) through the `--editorconfig` option. It will overide whatever other settings you have before the config file.

I need more comments and suggestions on how I shall proceed from here.

**Updated on 14 July 2016**
Added the feature to enable setting options from [Editorconfig](http://editorconfig.org/).
To enable this feature, simply add the option `--editorconfig` in your command, e.g.
```bash
    js-beautify --editorconfig your_script.js
```
This will then, for each filename you invoke `js-beautify` with, search for the respective settings from EditorConfig settings. This will overide all the options you supplied directly from the command line. Currently it only supports `indent-style` and `indent-size` and `max-line-length`.

The reason that you do not supply an `.editorconfig` file direcly is that EditorConfig itself maintains a mechanism to search and combine your hierarchy of `.editorconfig` files. It does not make sense and is cumbersome to supply your `.editorconfig` directly.


js/lib/cli.js
```diff
@@ -165,6 +168,34 @@ function getUserHome() {
...
+        var configs = configPath.map(function(path) {
+            return {
+                name: path,
+                contents: fs.readFileSync(path, 'utf8')
+            };
+        });
+        var eConfigs = editorconfig.parseFromFilesSync(opts.files[0], configs);
+        var _translation = {
+            "indent-size": "indent-size",
+            "max-line-length": "wrap-line-length"
```
bitwiseman: Why the difference between the names here?

ethanluoyc: its to translate the config names to one recognized by js-beautify

bitwiseman:
Looks reasonable so far. I had that one question.  
And also you need to add tests and the python translation. 
What other feedback would you like?

ethanluoyc:
Ok, I think I will soon add the python implementation and tests.

Sent from my iPhone, sorry for the brevity.




On Tue, Jul 12, 2016 at 2:01 PM +0800, "Liam Newman" <notifications@github.com> wrote:












Looks reasonable so far. I had that one question.


And also you need to add tests and the python translation. 

What other feedback would you like?

ethanluoyc:
@bitwiseman I noticed that the python version does not seem to yet support using a config file (like `jsbeautifyrc`, maybe we defer the implementation in js for now?

bitwiseman:
Unfortunately, "deferral for now" in this project means "deferral until bitwiseman does it" and these days I do not have the time. 

The best solution I've found is require python/javascript parity for all new commits, even when that means implementing a missing feature.  If you'd like to implement `editorconfig` support without `.jsbeatifyrc` that is fine, but I can't accept features that don't maintain parity.

ethanluoyc:
Ok, then I will try to implement in Python as soon as I can for the editorconfig part :) But for configuration with `.jsbeautifyrc` in Python I won't do

ethanluoyc:
I just added Python support, and will include a short write up explaining the newly added feature.

bitwiseman:
Sounds good.  Thanks again!

ethanluoyc:
Fixed compatibility issues between Py2 and Py3, should be good to go

bitwiseman:
Everything passes and you've added tests.  Please add a section to the README.md and I'll merge.

ethanluoyc:
Done. I also squashed the commits for easier merging :)

mitermayer:
Awesome!

js/lib/cli.js
```diff
@@ -165,6 +168,24 @@ function getUserHome() {
     return user_home;
 }
 
+function configEditorConfig(file) {
+    var eConfigs = editorconfig.parse(file);
+    var _translation = {
+        "indent-size": "indent-size",
+        "max-line-length": "wrap-line-length"
```
bitwiseman: Please add translation: `end_of_line` as `eol`.

js/lib/cli.js
```diff
@@ -326,6 +349,12 @@ function processInputSync(filepath) {
 
 function makePretty(code, config, outfile, callback) {
     try {
+        if (config.editorConfig) {
```
bitwiseman: I think putting  this here, will result in editorconfig values overriding parameters passed on the command line.  I would expect command line values to take precedence - editorconfig values are the default, then `.jsbeautifyrc` (if supported), and finally CLI parameters.  

And then tests need to enforce this.

python/setup.py
```diff
@@ -32,7 +32,7 @@ def run_tests(self):
       scripts=['js-beautify'],
       packages=['jsbeautifier', 'jsbeautifier.tests', 'jsbeautifier.tests.generated',
                 'jsbeautifier.unpackers', 'jsbeautifier.unpackers.tests'],
-      install_requires=["six>=1.6.1"],
+      install_requires=["six>=1.6.1", "editorconfig"],
```
bitwiseman: specify editorconfig version ">=" just to be safe.

bitwiseman:
Hey, added a few more comments.  I just want to say again - thank you, great work, and I'm eager to merge this new feature.  :smile:

js/lib/cli.js
ethanluoyc: I was trying to do this but I had some difficulties. This is because you have to obtain the editorconfig settings for each individual file, that is after the parsing of all the configs but before the prettifier actually starts to work. It is hard to override the settings at that time since there is no way to detect if the configuration is set by a default value.

ethanluoyc: @bitwiseman https://github.com/beautify-web/js-beautify/blob/master/js/lib/cli.js#L182 starting here, the configuration is set up prior to inspecting into each individual file. I cannot hook up the editorconfig configs here since I have no idea about the specific one file that will be beautified. But I also cannot hook up right before makePretty since the default settings already kick in. Unless we rewire the logic of the config step I doubt if there is a simple way to allow individual settings to override editorconfig easily

bitwiseman: @ethanluoyc - I see what you're saying.  The alternative is to say that editorconfig values superceed all other config values.  That could be acceptable - the user is specifying "Use EditorConfig". 

We still need tests that show and enforce that behavior.

ethanluoyc:
@bitwiseman make the modification as you specified. Tests have been included here https://github.com/beautify-web/js-beautify/pull/969/commits/d2861bd1eb2ed5d7be05b19de304db26187c3ac5#diff-01f1e754995e6be926872bf107bcb867R115 Not sure if this will suffice.

bitwiseman:
@ethanluoyc - Squashed your changes together and then added some more.  Merging your changes shortly.

maksimr:
@ethanluoyc thanks for this PR but now it is useless if you don't pass file path.
Many editors does not pass file path because it's allow format only part of file, so it will be cool add functionality which allow for example pass path to .editorconfig file from which we can read properties

Thanks!

bitwiseman:
@maksimr - please file a new issue.

PR #985, opened 2016-08-04
CSS: Preserve white space before pseudo-class and pseudo-element selectors

Konamiman:
This issue was raised in [a question in Meta Stack Overflow ](http://meta.stackoverflow.com/questions/309573/another-bug-in-snippets-tidy-button) - Stack Overflow uses js-beautify in the code snippets editor.

As explained in the question, the CSS beautifier is converting `tr :first-child {...}` into `tr:first-child {...}`, which is incorrect since both things don't mean the same. This change fixes this.


Konamiman:
I have changed the `beautify-css.js` file and added the appropriate js tests. Python tests have been automatically added but they fail. Why?

HookyQR:
@Konamiman `div:focus` and `div :focus` are not the same thing.

`div:focus` styles are activated when a div has the focus.
`div :focus` styles are activated when a **child of** div has the focus.

bitwiseman:
Ah, you need to update the python implementation as well. It is really easy. 
https://github.com/beautify-web/js-beautify/blob/master/python/cssbeautifier/__init__.py

python/cssbeautifier/__init__.py
```diff
@@ -409,6 +409,10 @@ def beautify(self):
                 else:
                     # sass/less parent reference don't use a space
                     # sass nested pseudo-class don't use a space
+
+					# preserve space before pseudoclasses/pseudoelements, as it means "in any child"
```
bitwiseman: Alignment.

bitwiseman:
Alignment of the comment in the python code.  Then I can merge!

Konamiman:
@bitwiseman Ok, done. Python is really tricky with that tabs vs spaces thing. Thank you!

bitwiseman:
https://github.com/Glavin001/atom-beautify/issues/1158

PR #999, opened 2016-08-18
Try to fix space_or_wrap method

### LINKED ISSUE #837: HTML wrap length line cut-off is problematic
HTML cut off is longer than the character count specified in settings. Example:

I have this segment in a file. I set the wrap line length to 80 (same as my editor line) and the result after beautifying is this:

```
  <div autocomplete placeholder="Type to select a company or create a new one*" class="search-container search-filter-container company-search"
  ng-class="{'for-new-company': modal.newCompany}" search-model="modal.companySearch"
  results="modal.companies" result-label="result.name" model-options="modal.searchOptions"
  createable="true" createable-condition="!modal.exactCompanyFound(searchModel) && (modal.step !== 'new-company' || modal.companySearch !== modal.newCompany.name)"
  createable-label='Create company "{{modal.companySearch}}"' on-select="modal.selectCompany(result)"
  on-create="modal.createCompany(searchModel)"></div>
```

Longest line is at about 144.

When I set the wrap line length to 40, I get this:

```
 <div autocomplete placeholder="Type to select a company or create a new one*"
  class="search-container search-filter-container company-search"
  ng-class="{'for-new-company': modal.newCompany}"
  search-model="modal.companySearch" results="modal.companies"
  result-label="result.name" model-options="modal.searchOptions"
  createable="true" createable-condition="!modal.exactCompanyFound(searchModel) && (modal.step !== 'new-company' || modal.companySearch !== modal.newCompany.name)"
  createable-label='Create company "{{modal.companySearch}}"'
  on-select="modal.selectCompany(result)"
  on-create="modal.createCompany(searchModel)"></div>
```

Seems to me that the check for wrapping is only when the attribute BEGINS after the specified amount - I would expect an alternate mode where any attribute that ENDS after the length will try to cut off, if it's too long on its own, it should cut itself off and then be in its own line that is longer and will be wrapped.

So evidently, I would like this result:

```
<div autocomplete
  placeholder="Type to select a company or create a new one*"
  class="search-container search-filter-container company-search"
  ng-class="{'for-new-company': modal.newCompany}"
  search-model="modal.companySearch"
  results="modal.companies"
  result-label="result.name"
  model-options="modal.searchOptions"
  createable="true"
  createable-condition="!modal.exactCompanyFound(searchModel) && (modal.step !== 'new-company' || modal.companySearch !== modal.newCompany.name)"
  createable-label='Create company "{{modal.companySearch}}"'
  on-select="modal.selectCompany(result)"
  on-create="modal.createCompany(searchModel)"></div>
```

Which thinking of it, can be reduced to another type of behavior: I know you can set `"wrap_attributes"` to be `auto` or `force`. `force` is closest to what I want, but the problem is that it forces even small elements with 2-3 attributes, which I'd rather keep in one line. I'm thinking there should be a third option, that will work like `force`, but only in the case of over `x` (let's say 5?) attributes on the same element. 

Are one of these possible? Thank you!

rodfersou:
Try to fix #837

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -234,7 +234,8 @@
             // at the wrap_line_length, append a newline/indentation.
             // return true if a newline was added, false if a space was added
             this.space_or_wrap = function(content) {
-                if (this.line_char_count >= this.wrap_line_length) { //insert a line when the wrap_line_length is reached
+                var next_token = this.input.slice(this.pos - 1).split(' ')[0];
+                if ((content + next_token).length >= this.wrap_line_length) { //insert a line when the wrap_line_length is reached
                     this.print_newline(false, content);
                     this.print_indentation(content);
                     return true;


bitwiseman:
Please add tests and run `./build` to verify your added functionality.

rodfersou:
after this change, many tests fail, and I don't have time now to fix all tests..
what do you think about the implementation? is this the right way to do it?

bitwiseman:
@rodfersou 
`get_token` advances the position of parsing and changes state.  What you need to make this work is a `peek_token`.  You could do that in the current design but it seems like it would be error prone.

What we really need to do is parse the input into a list of tokens before we try to fomat them, unlike now where we parse as we go.

rodfersou:
@bitwiseman 
agree.. this parser works in a difficult way to maintain, and it need a complete rewrite.. unfortunately I don't have time to do it..

bitwiseman:
@rodfersou - Yeah, rewrite.  I've been rewriting the supporting layers of the javascript beautifier, believe me I know how slow going it can be.  Thanks for trying.

rodfersou:
you are welcome

PR #1025, opened 2016-09-08
Options for css and js inside html files

DaniGuardiola:
This should be useful for plugins and for developers looking for a more powerful beautifying tool for html


js/lib/beautify-html.js
```diff
@@ -84,7 +84,7 @@
         return s.replace(/\s+$/g, '');
     }
 
-    function style_html(html_source, options, js_beautify, css_beautify) {
+    function style_html(html_source, options, js_beautify_raw, css_beautify_raw) {
```
bitwiseman: Instead of changing this, why not use different names for the `js-beautify` and `css-beautify` vars. 😄

js/lib/beautify-html.js
```diff
@@ -1034,4 +1052,4 @@
         };
     }
 
-}());
\ No newline at end of file
+}());
```
bitwiseman: You must run `./build` to ensure your code meets style guidelines for the project.

bitwiseman:
@DaniGuardiola - I'm glad to see someone working on this.  Thanks!

js/lib/beautify-html.js
bitwiseman: But more accurately, I think you'd do better to change each beautifier so each on reads options + options.<language>.  Other wise you only solve this for the case where people are running the html beautifier.

js/lib/beautify-html.js
```diff
@@ -108,6 +108,24 @@
 
         options = options || {};
 
+        // Allow the inclusion of css and js options for <style> and <script> blocks
+        // inside html files (this is useful for plugins like Sublime's HTML prettify)
+        if (!options.js) {
+          options.js = {};
+        }
+        var js_beautify = function(js_source_text, options) {
+            options = options || options.js;
```
bitwiseman: This is an interesting start, but what you really want is something that overlays any values in `options.js` onto `options` object. 

And don't forget to do the same with `options.html`.

So, for instance your the following settings should work and result in html files with 8 space indents, `<style>` blocks wth 4 space indent, and `<script>` blocks with 2-space indents. 

```json
{
    "indent_size": 5,
    js: {
        "indent_size": 2
    },
    html: {
        "indent_size": 8
    }
}
```

js/lib/beautify-html.js
DaniGuardiola: Well I don't think anything similar to this will be useful for beautifying js or css as those files will never have blocks with the other two languages as html does

DaniGuardiola:
idk why the tests fail but this should be working just fine now

js/lib/beautify-html.js
bitwiseman: But they can all use the same configuration file or options object.  So the same schema needs to work for all of them.

bitwiseman:
@DaniGuardiola - Great!  What you have doesn't break the existing tests, now we need tests showing the new feature working.  Thanks!

HookyQR:
I would prefer to see a merge of the options at the embedded level, so things don't have to be defined twice. So that only *specifically different* settings need to be defined at the embedded level(s).

For instance:
```javascript
{
    "indent_with_tabs": false,
    "brace_style": "collapse",
    "max_preserve_newlines": 2,
    "js":{
        "indent_with_tabs": true
    }
}
```

The javascript beautify would only get the `indent_with_tabs` setting with the current code, and drop the `brace_style` and `max_preserve_newlines`.


So instead of calling `opt = options.js || opt` call `opt = mergeOpts(opt, "js")`

```javascript
function mergeOpts(allOptions, targetType) {
	var finalOpts = {}, name;
	for (name in allOptions) {
		if (name !== 'js' && name !== 'html' && name !== 'css') {
			finalOpts[name] = allOptions[name];
		}
	}
	//merge in the per type settings for the targetType
	if (targetType in allOptions) {
		for (name in allOptions[targetType]) {
			finalOpts[name] = allOptions[targetType][name];
		}
	}
	return finalOpts;
}
```

This is the merge I'm using in [VSCode Beautify](https://github.com/HookyQR/VSCodeBeautify).

HookyQR:
The function overwrite also circumvents these tests, and will break execution if, for some reason, the import(s) failed.
```javascript
if (multi_parser.token_type === 'TK_SCRIPT') {
    _beautifier = typeof js_beautify === 'function' && js_beautify;
} else if (multi_parser.token_type === 'TK_STYLE') {
    _beautifier = typeof css_beautify === 'function' && css_beautify;
}
```

bitwiseman:
@HookyQR - What does your second comment there mean?

bitwiseman:
Taking the code from this and integrating elsewhere.

DaniGuardiola:
@bitwiseman hey sorry, this was on my to-do list for a long time, can I help?

bitwiseman:
@DaniGuardiola @HookyQR  - You are welcome to look at the recently merged commits, sync master, and see if the behavior suits your needs.

https://github.com/beautify-web/js-beautify/commit/1c21d5b99af4298d82c9b8b5a746de82e476e98f

It was a bit more involved to get it wired through and tested throughout.

PR #1070, opened 2016-12-09
New option "force-expand-multiline" for "wrap_attributes"

tobias-zucali:
This adds a new option `"force-all"` for `"wrap_attributes"`.

- if there is just one attribute, it is not wrapped (like `"wrap_attributes": "force"`
- if there is more than one attribute, each of them has it's own line
- if attributes are wrapped, the ending part of the tag (`>` or `/>`) is in its own line

For example:
``` html
<a
  accesskey="l"
  class="link link-blue"
  href="https://github.com/beautify-web/js-beautify/"
>
  Link to a great <strong class="link-emphasis">project</strong>
</a>
```

We use this formatting style because
- all attributes are aligned without losing the indent from nesting level like with `force-aligned`
- attributes can be easily sorted (we use alphabetical sorting for attributes, would be great to implement that in js-beautify as well)
- the ending part of the tag on it's own line emphasises the indent of the tag and makes indent level more visible

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ HTML Beautifier Options:
   -b, --brace-style                  [collapse-preserve-inline|collapse|expand|end-expand|none] ["collapse"]
   -S, --indent-scripts               [keep|separate|normal] ["normal"]
   -w, --wrap-line-length             Maximum characters per line (0 disables) [250]
-  -A, --wrap-attributes              Wrap attributes to new lines [auto|force|force-aligned|force-all] ["auto"]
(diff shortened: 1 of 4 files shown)


bitwiseman:
Wow,  great PR.  Tests and everything.  Thanks!

Only one change needed: the name `force-all` needs changing. `force-all` indicates that you're going to "force all" items to do something, which is not the case.  How about `force-expand-multiline`?

tobias-zucali:
@bitwiseman I changed `force-all` to `force-expand-multiline`, hope it is ok now. Just tell me if I can improve something else!
I am happy that I was able to contribute ;-)

blemaire:
This feature is currently broken #1335

PR #1104, opened 2017-01-06
Ignore linebreak when meet handlebar

mmsqe:
Add test


mmsqe:
Prevent line break by options of wrap line length within handlebars bracket. Before behavior is that statement will be broken when reach the max line length like below.
```
<span>{{condition1 && condition2 && condition3 && 
	condition4 ? "resForTrue" : "resForFalse"}}</span>
```

PR #1131, opened 2017-02-10
Fix #1102 and fix #1103

### LINKED ISSUE #1102: force-aligned is not aligned when indenting with tabs
# Description
Force aligned uses the indent character to provide the _correct_ alignment. If the indent character is set to '\t' or indent-with-tabs is used, the result is wrong.

# Input
```html
<div first second></div>
```

# Expected Output
```html
<div first
     second></div>
```

# Actual Output
```html
<div first
					second></div>
```

# Steps to Reproduce

## Environment
OS: MacOS

## Settings
Example:
```json
{
    "indent_with_tabs": true,
    "wrap_attributes": "force-aligned"
}
```

### LINKED ISSUE #1103: Lines are not un-indented correctly when attributes are wrapped
# Description
Indenting is broken whenever attribute wrapping happens.

Using js-beautify **v1.6.8**.

# Input
```html
<div first second>content</div>
<div>content</div>
```

# Expected Output
```html
<div first 
     second>content</div>
<div>content</div>
```

# Actual Output
```html
<div first
     second>content</div>
    <div>content</div>
```

# Steps to Reproduce
Beautify the code

## Environment
OS: MacOS

## Settings
Example:
```json
{
    "wrap_attributes": "force-aligned"
}
```
**Note:** wrap_attributes doesn't have to be set to force-aligned. Any wrapping (excluding "force-expand-multiline") introduces, including if "auto" caused the wrapping will demonstrate this issue.

HookyQR:
Fixes attribute wrapping issues.

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -490,7 +490,8 @@
                             }
 
                             for (var count = 0; count < alignment_size; count++) {
-                                content.push(indent_character);
+                                // only ever further indent with spaces since we're trying to align characters
+                                content.push(' ');
                             }
                         }
                         if (first_attr) {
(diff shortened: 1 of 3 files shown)


rosay:
Thanks for the fix, @HookyQR!

PR #1143, opened 2017-02-22
Fix preserve_newlines issue

HookyQR:


HookyQR:
The revert will have to be dropped for this merge to work. (Or manual merge.) This PR based off the original `preserve_newlines` PR.

test/data/css/tests.js
```diff
@@ -230,6 +230,19 @@ exports.test_data = {
...
         ],
     }, {
+        name: "Preserve Newlines and newline_between_rules",
+        description: "",
+        options: [
+            { name: "preserve_newlines", value: "true" },
+            { name: "newline_between_rules", value: "true" },
+        ],
+        tests: [
+            { input: '.div {}.span {}', output: '.div {}\n\n.span {}' },
```
bitwiseman: Take these tests and move them into the previous section. 
Then change the matrix options above to  the 4 combinations of `preserve_newlines` and `newline_between_rules`.

test/data/css/tests.js
```diff
@@ -230,6 +230,19 @@ exports.test_data = {
             { input: '#bla, #foo{\n\tcolor:black;{{separator_input}}\tfont-size: 12px;\n}', output: '#bla,\n#foo {\n\tcolor: black;{{separator_output}}\tfont-size: 12px;\n}' }
```
bitwiseman: The output here has `bla,\n#foo`.  We need to preserve newlines here, too.  So need inputs such as `#bla\n,\n\n#foo` - The newline before the comma should go away, but the other two?

test/data/css/tests.js
```diff
@@ -230,6 +230,19 @@ exports.test_data = {
...
+        name: "Preserve Newlines and newline_between_rules",
+        description: "",
+        options: [
+            { name: "preserve_newlines", value: "true" },
+            { name: "newline_between_rules", value: "true" },
+        ],
+        tests: [
+            { input: '.div {}.span {}', output: '.div {}\n\n.span {}' },
+            { unchanged: '.div {}\n\n.span {}' },
+            { unchanged: '.div {\n\ta: 1;\n\n\n\tb: 2;\n}\n\n\n\n.span {\n\ta: 1;\n}' },
```
bitwiseman: What about `'.div {\n\n\n\ta:`?

bitwiseman:
Thanks for the fast turn around.  Could you do the `git revert  ba63466` as  the first commit of this PR?   Then it will merge. 
Also, we need to see tests of newline chars in a number of other places, as noted.

test/data/css/tests.js
HookyQR: They don't merge up well because of exactly the issue that requires this fix. I'll add ones from above here.

HookyQR:
Rebased and added tests per request.

test/data/css/tests.js
HookyQR: It seems the original code didn't preserve new lines there anyway. It only preserves new lines between, and within rules, not in the selector list.

test/data/css/tests.js
HookyQR: Added

HookyQR:
Closing this. There is another issue with the preserve new line code. It can remove selectors completely. It just so happens to be hackdays at work today. I believe I know what I'll be doing. ☹️ 😉

PR #1145, opened 2017-02-22
No extra newline following header comment in CSS

### LINKED ISSUE #531: Extra newline is inserted after the comment line instead of before it

joelwilliamson:
Add a cli flag to control whether CSS  block comments in the header
are followed by a blank line.

This is intended to fix #531 . I would be happier not having it controlled by a switch or defaulting to no newline, but am open to advice

### DIFF
--- a/js/lib/beautify-css.js
+++ b/js/lib/beautify-css.js
@@ -104,6 +104,7 @@
         var space_around_combinator = (options.space_around_combinator === undefined) ? false : options.space_around_combinator;
         space_around_combinator = space_around_combinator || ((options.space_around_selector_separator === undefined) ? false : options.space_around_selector_separator);
         var eol = options.eol ? options.eol : 'auto';
+        var newline_following_header = (options.newline_following_header == undefined) ? true : options.newline_following_header;
 
         if (options.indent_with_tabs) {
             indentCharacter = '\t';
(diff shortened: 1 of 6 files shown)


bitwiseman:
@joelwilliamson 
Sorry for slow response.  Please resolve the conflicts since you submitted.  
Also, since this is intended to fix #531, please include the same input from #531 in your tests.  

I don't think you need to make this a new option.  Does making it the standard break any existing tests?  If not, then certainly make this the default behavior.  If it does break tests, take a look at the way they break and consider whether the test is wrong.  

Thanks!

bitwiseman:
@joelwilliamson - any chance to work on this?

Boycce:
Does this solve #1219 ?

bitwiseman:
@Boycce - don't know if this fixes #1185.
@joelwilliamson - If you have time to work on this, it would be useful.

Anyone else wanting to take up Joel's work is also welcome.

bitwiseman:
@joelwilliamson - If you have time to work on this, it would be useful.

Anyone else wanting to take up Joel's work is also welcome.

PR #1146, opened 2017-02-24
(re)add preserve_newlines for css

HookyQR:


HookyQR:
This is ready for review. It implements the preserve_newlines for css, but doesn't break anything else.

js/lib/beautify-css.js
```diff
@@ -234,6 +241,16 @@
             return false;
         }
 
+        function removeWhiteSpaceOnEmptyLines(input) {
```
HookyQR: I don't think this actually has any effect the way the newlines are handled now. It's highly likely this is an expensive `noop`. I'll take a look tonight.

HookyQR:
No problems @bitwiseman. The code as it was, basically ignored whitespace and put in what it wanted to. Changing it to preserve white space means it actually has to do something with the existing whitespace. It's definitely the right place to handle it, but it now effectively makes the method name wrong. The real problem lies in the fact that (as I'm pretty sure you've mentioned before) the processor doesn't have a parser, it's effectively a streaming filter.

I'm aware this is a big change in the way the system functions, but I know it works (does what the original change said it was going to do). I'm glad you've pulled in others to test the functionality though.

I'd like to refactor the whole thing to do parse and rebuild, which would allow a substantial amount more functionality - some of which people have been asking for from my extension, which uses this module. Unfortunately, like you, I simply don't have the time.

djayard:
@HookyQR @bitwiseman 
It's working for the cases that caused me to raise an issue; the beautifier is consistent and is no longer generating new lines between CSS rule sets in html style elements.

HookyQR:
@bitwiseman I assume you're not merging this because of your refactor work. Is that right?

bitwiseman:
@HookyQR - Nope, I've been offline for about a week and then busy.  I'll merge this this coming weekend.

octref:
Waiting for this one to land so I can include it in my vue extension :D

HookyQR:
Will we get a new release for this @bitwiseman?

bitwiseman:
Done.

PR #1148, opened 2017-02-27
made unformatted options recognize namespaced XML tags

jorahood:


js/lib/beautify-html.js
```diff
@@ -984,7 +984,7 @@
                     break;
                 case 'TK_TAG_SINGLE':
                     // Don't add a newline before elements that should remain unformatted.
-                    var tag_check = multi_parser.token_text.match(/^\s*<([a-z-]+)/i);
+                    var tag_check = multi_parser.token_text.match(/^\s*<([a-z:-]+)/i);
```
bitwiseman: This would allow elements to be `a:b:c:d`, which is not valid.

test/data/html/tests.js
```diff
@@ -761,8 +761,24 @@ exports.test_data = {
             { fragment: true, unchanged: '<div class="searchform"><input type="text" value="" name="s" id="s"><input type="submit" id="searchsubmit" value="Search"></div>' },
         ]
     }, {
+        name: "unformatted tags with namespaces",
+        description: "namespaced tags with colons can be specified for leaving unformatted",
+        options: [ {name: 'unformatted', value: '[ "name:space"]'}],
+        tests: [{
```
bitwiseman: Please add more tests, including some tests that show invalid elements such as `a:b:c`
Also, need to support and test these elements when they are not unformatted.

bitwiseman:
@jorahood - Any update on this?

bitwiseman:
@jorahood - this seems like a worthwhile change. I just needs some tweaking.  Why close?

Issue #1153, opened 2017-03-07
Newlines are inserted between anchor tags around images


EphoxJames:
# Input
The code looked like this before beautification:
```html
<p><a href="#"><img src="https://1k96m612hyol3k0gtk3u20vw-wpengine.netdna-ssl.com/wp-content/uploads/2017/01/tinymce-cloud-header@2x-600x400.jpg" style="width:100px"/></a><a href="#"><img src="https://1k96m612hyol3k0gtk3u20vw-wpengine.netdna-ssl.com/wp-content/uploads/2017/01/tinymce-cloud-header@2x-600x400.jpg" style="width:100px"/></a></p>
```
Note that there are no spaces between the two images despite them being wrapped in anchor tags.

# Expected Output
Spaces should not have been put between the two anchor tags because it changes the way the HTML is rendered (otherwise I don't mind).

This is one option for what the code could look like after beautification:
```html
<p>
  <a href="#"><img src="https://1k96m612hyol3k0gtk3u20vw-wpengine.netdna-ssl.com/wp-content/uploads/2017/01/tinymce-cloud-header@2x-600x400.jpg" style="width:100px" /></a><a href="#"><img src="https://1k96m612hyol3k0gtk3u20vw-wpengine.netdna-ssl.com/wp-content/uploads/2017/01/tinymce-cloud-header@2x-600x400.jpg" style="width:100px" /></a>
</p>
```

# Actual Output
Spaces were inserted between the two anchor tags which results in a space being rendered
between the two images when the HTML is displayed.

The  code actually looked like this after beautification:
```html
<p>
  <a href="#"><img src="https://1k96m612hyol3k0gtk3u20vw-wpengine.netdna-ssl.com/wp-content/uploads/2017/01/tinymce-cloud-header@2x-600x400.jpg" style="width:100px" /></a>
  <a href="#"><img src="https://1k96m612hyol3k0gtk3u20vw-wpengine.netdna-ssl.com/wp-content/uploads/2017/01/tinymce-cloud-header@2x-600x400.jpg" style="width:100px" /></a>
</p>
```

# Steps to Reproduce
Can be reproduced on the website http://jsbeautifier.org/ by pasting in the above input and running the beautifier.

## Environment
OS: Linux Mint
Browser: Firefox

## Settings
```json
{
        "indent_size": "2",
        "indent_char": " ",
        "max_preserve_newlines": "5",
        "preserve_newlines": true,
        "keep_array_indentation": false,
[...]

PR #1158, opened 2017-03-14
Fix for issue 1153: Newlines are inserted between anchor tags around images

jhaines:
The tag variable in this function is presently an array, returned from the match called on the next tag.

Since the Utils.in_array function called on the list of unformatted tags requires a string, we need the second value of the array, which returns the subgroup with the tag name, such as "img" in "\.


jhaines:
I've added some more tests.  Also one of the previous tests had a typo of "fragement" instead of "fragment".  I don't think it was causing issues, but I've fixed it regardless.

TheSpyder:
Github hasn't linked them due to the lack of a # in the title, but this fixes #1153

bitwiseman:
@jhaines - thanks, I'm going to wait a few more days on to merge this, but it looks good to me.

TheSpyder:
Has enough time passed? This triggers a minor bug in our editor and it would be nice to have an updated JS Beautifier release to fix it :)

PR #1169, opened 2017-03-28
Fix label checking incorrect box when clicked

Smashman:
I noticed that "Preserve inline braces/code blocks?" label was ticking "Keep array indentation?" checkbox


Smashman:
Soooo...?

Issue #1170, opened 2017-03-31
Invalid attribute name should be ignored


Janpot:
# Description

Let me start of by noting that this issue is about invalid html. It's a bit of an edge case but let me explain how I obtain this html. I'm trying to beautify a html string that comes from a DOM element that I don't control. I use `element.innerHTML` and feed the result into the html beautifier. Unfortunately some invalid input still produces a working DOM. example:

```js
new DOMParser().parseFromString(`<div "><div>hello</div></div>`, 'text/html').body.innerHTML
// <div "=""><div>hello</div></div>
```

If I understand correctly, parsers should just ignore invalid attribute names and continue parsing.

# Input
The code looked like this before beautification:
```
<div "=""><div>hello</div></div>

<div "><div>hello</div></div>
```

# Expected Output
The code should have looked like this after beautification:
```
<div "="">
  <div>hello</div>
</div>

<div ">
  <div>hello</div>
</div>
```

# Actual Output
The  code actually looked like this after beautification:
```
<div "=" "><div>hello</div></div>

<div "><div>hello</div></div>
```

## Settings
Default

Janpot:
ever thought about using [htmlparser2](https://www.npmjs.com/package/htmlparser2)?

bitwiseman:
@Janpot - We've tried to use minimal node-specific dependencies in order to support both javascript and python implementations with as much of the codebase being the same as possible.

PR #1171, opened 2017-03-31
Handle quotes as attributes

Janpot:
Fixes https://github.com/beautify-web/js-beautify/issues/1170


bitwiseman:
I see what you are doing here, but I don't really trust it sorry. 
This is the kind of change that could break in a number of unexpected ways (ways I can't even think of off the top of my head).  

I'm going to need you to put on your testing hat (or find someone who you trust to do so) and add  something like 10 to 20 more tests to this.   It is not enough to show that this works in the cases you know, I need to see examples of it not breaking under pressure.

Janpot:
@bitwiseman After doing this PR and digging around the code I figured the way the parser is built is too unreliable for me. In my application the output of this module was used for further processing and in my latest version I completely removed it. I wouldn't mind so much if the use was purely for cosmetic purpose. I will close this PR as I don't have any use in spending a lot of time on this anymore.
My advice would be to build this beautifier on top of a reliable, forgiving and well tested html parser.

P.S. I don't want to come of as too negative, I think this is a good and useful module. Just not forgiving enough towards certain use cases, like mine.

PR #1176, opened 2017-04-11
addOptionsExplicationSection

samuelt1:
I added a section to the readme file that allows people with no understanding of the project to quickly understand what each of the options do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ var a =  1;
 The number of characters to use for indentation.
 
 ## indent-char
-The acutal character that will be used to indent your code. It can be anything.
+The actual character that will be used to indent your code. It can be anything.
 
 ## indent-with-tabs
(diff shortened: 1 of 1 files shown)


samuelt1:
> Start each description with text from options output. Then add whatever other comments you feel are needed.
what do you mean by this?

This line would look like this?
```
## indent-size
Indentation size [4]
The number of characters to use for indentation.
```

I am not sure what the HTML and CSS does, so I cant really edit those.

and why do you want it in alphabetical order? I put it in the order it appears in the 

> Beautifier Options: 

section

bitwiseman:
@samuelt1 
Yes, the example you showed is what I intended. 

I suggested alphabetical order so they would be easy for people to lookup.

bitwiseman:
@samuelt1  Was this done?

PR #1191, opened 2017-05-11
Add initial indent level

aeschli:
Allows specifying the initial indentation level to be used. Typically when formatting a full document that is 0, but when formatting a snippet of HTML that should be inserted in an existing document or formatting a range of a document ('format selection') it is useful to be able to get an initial indent greater than 0.

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -890,7 +890,7 @@
                         return;
                     }
                     if (text || text !== '') {
-                        if (this.output.length && this.output[this.output.length - 1] === '\n') {
+                        if (!this.output.length || this.output[this.output.length - 1] === '\n') {
                             this.print_indentation(this.output);
                             text = ltrim(text);
                         }
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -43,6 +43,7 @@
     indent_inner_html (default false)  — indent <head> and <body> sections,
     indent_size (default 4)          — indentation size,
     indent_char (default space)      — character to indent with,
+    indent_level (default 0)         — the initial indent level,
     wrap_line_length (default 250)            -  maximum amount of characters per line (0 = disable)
     brace_style (default "collapse") - "collapse" | "expand" | "end-expand" | "none"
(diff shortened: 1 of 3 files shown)


bitwiseman:
@aeschli - 
Thanks for the PR! 

The javascript and css beautifiers do this automatically by retaining the whitespace at the start of the input string as the base indent.   I'm surprised the html doesn't do the same, so thanks for drawing attention to it.  Perhaps you could port the javascript functionality instead?   

See `baseIndentString` or `preindent_index`:
https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L392 

Either way, please add tests as well, and then I can merge.

aeschli:
@bitwiseman Thanks for the quick reply! I added a test and even found an issue.

'Preserve indent' is less useful for my use case as I would have to add an artificial indent to my snippet. But I fully understand if you prefer that over my suggestion.

bitwiseman:
@aeschli  
If you're going to add this option, I'd like to see it work for all three beautifiers.  "Preserve indent" is less useful to you, but it would be useful to others.  Also, by porting the "preserve indent" functionality to the html beautifier, you could then build the `indent_level` on top of it for all the beautifiers.   

Please have `indent_level` default to `auto`.   Whether you want to port "preserve indent" or not, having the default be non-numeric will make the option you're adding compatible with the existing functionality.   Setting `indent_level` would then override the "preserve indent" and set the starting indent level to that value, with `0` being no indent. 

The one piece of added complexity I can see from adding the option to all three beautifiers, is that we'd need to modify the setting when passing inherited options from the html beautifier into the js and css beautifiers.

I understand if you don't want to do this - It takes you minor change to quite a bit more.  I'll take your   But as long as you're contributing, it would be really great to have 

In any case, for tests - please add tests for the following values: `-1`, `0`, `4`, and `auto`.

bitwiseman:
@aeschli -  Updates?

aeschli:
Sorry, I realize that this change has more ripples than I thought, I didn't think of JS and CSS. As I have an easy workaround (add indentation on the formatted result), so let's avoid the extra option. Closing the PR.

Issue #365, opened 2013-12-05
Vertically align (indent) values in object literals


binarykitchen:
Would be nice to modify from

``` js
{
    browserName: 'internet explorer',
    platform: 'Windows 7',
    version: '8'
}
```

to 

``` js
{
    browserName: 'internet explorer',
    platform:    'Windows 7',
    version:     '8'
}
```

but cannot find such an option. Any clues? Or is this functionality missing?

bitwiseman:
This is getting closer to being possible.  We actually do properly detect object blocks now.

DawidaWina:
Will you do something with that topic? I can see that this issue is still exists almost 3 years now and my ticket was closed. I don't know if I should wait or try to rewrite it and send pull request of my solution. :)

evocateur:
Pull requests are always welcome.

gentlefox:
Possible solution (unsure how Q&D attempt is going - I'll try it out later).
The [Alignment](https://github.com/wbond/sublime_alignment) plugin (Sublime Text, Textmate etc) utilises .jsbeautifyrc and offers the following settings and the desired functionality of this enhancement.
**Alignment Settings**

```
    {
      // If the indent level of a multi-line selection should be aligned
      "align_indent": true,

      // If indentation is done via tabs, set this to true to also align
      // mid-line characters via tabs. This may cause alignment issues when
      // viewing the file in an editor with different tab width settings. This
      // will also cause multi-character operators to be left-aligned to the
      // first character in the operator instead of the character from the
      // "alignment_chars" setting.
      "mid_line_tabs": false,

      // The mid-line characters to align in a multi-line selection, changing
      // this to an empty array will disable mid-line alignment
      "alignment_chars": ["=", ":"],

      // If the following character is matched for alignment, insert a space
      // before it in the final alignment
      "alignment_space_chars": ["=", ":"],

      // The characters to align along with "alignment_chars"
      // For instance if the = is to be aligned, there are a number of
      // symbols that can be combined with the = to make an operator, and all
      // of those must be kept next to the = for the operator to be parsed
      "alignment_prefix_chars": [
        "+", "-", "&", "|", "<", ">", "!", "~", "%", "/", "*", "."
      ]
    }
```

Alternative setting, following **.eslintrc** rules:

```
    'indent': ['error', 2, {
      'FunctionExpression': {
        'parameters': 'first'
      },
      'CallExpression': {
        'arguments': 'first'
      },
      'ArrayExpression': 'first',
[...]

Issue #396, opened 2014-01-31
Option to preserve existing var and object alignment (like preserve_array_indentation)


QuentinFchx:
Would it be possible/acceptable to add an option that converts

```
var short_var = "value";
var very_long_var = "value";

var obj = {
    "short_prop": "value",
    "very_long_property": "value" 
}
```

to

```
var short_var     = "value";
var very_long_var = "value";

var obj = {
    "short_prop"        : "value",
    "very_long_property": "value" 
}
```

Nikita240:
+1 PHP_Beautifier has this feature 

http://pear.php.net/reference/PHP_Beautifier-latest/PHP_Beautifier/Filter/PHP_Beautifier_Filter_EqualsAlign.html

bitwiseman:
I have no idea what bearing the features of PHP_Beautify has have on this project.  :smile:

Nikita240:
@bitwiseman It demonstrates a precedent for the feature

bitwiseman:
Sorry, I was mostly teasing you.  :smile: 
I already agreed this is a valid and potentially useful feature, just lower priority and high difficulty than other currently open issues.

Nikita240:
I'll try to find some free time to make a PR then :)

bitwiseman:
A little unsolicited advice: I'd try implementing this for just the object literal case first.  It's relatively cleanly recognized start and end, during which you can easily track the lines where fields are declared and the field lengths.  Do a PR for that, we can discuss and work out any rough edges, and the go back and do that var case next.   If you get stuck, drop me a line or an intermediate branch to look at.

ryanoasis:
Excuse my ignorance but is there any current way to preserve object formatting like:

```
var obj = { a: 123, b: "xyz" };
```

chiluap:
eslint can do this for you with 
~~~Fix all auto-fixable problems~~~
and the configuration:
~~~"key-spacing": [2, {
        "afterColon": true,
        "align": "value",
        "beforeColon": false
    }]~~~

clawconduce:
This should link to #365 and #603

bitwiseman:
@clawconduce - Agreed, thanks.

creativecag:
+1

clawconduce:
See #1209 , it's not pretty, but it's at least a start, and could be mergeable with some love.

clawconduce:
I gave this a quick go in #1209, but I don't think I'll have the time to properly wrap it up.

[...]

PR #1209, opened 2017-06-26
Quick and Dirty attempt at aligning object literals #365

clawconduce:
I should start by saying I'm not a fan of the code here, but I do like the output.

I want to get the ball rolling on #365 but I am definitely not familiar with the codebase, but my hope is that this will eventually lead to getting this issue resolved.

### DIFF
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -458,6 +458,7 @@ function Beautifier(js_source_text, options) {
     opt.jslint_happy = (options.jslint_happy === undefined) ? false : options.jslint_happy;
     opt.space_after_anon_function = (options.space_after_anon_function === undefined) ? false : options.space_after_anon_function;
     opt.keep_array_indentation = (options.keep_array_indentation === undefined) ? false : options.keep_array_indentation;
+    opt.align_object_literals = (options.align_object_literals === undefined) ? false : options.align_object_literals;
     opt.space_before_conditional = (options.space_before_conditional === undefined) ? true : options.space_before_conditional;
     opt.unescape_strings = (options.unescape_strings === undefined) ? false : options.unescape_strings;
     opt.wrap_line_length = (options.wrap_line_length === undefined) ? 0 : parseInt(options.wrap_line_length, 10);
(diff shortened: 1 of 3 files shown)


clawconduce:
I'm not sure why AppVeyor failed.

I wanted to note that I just saw that Python changes are required as well.  I will port it over once the approach looks like it's reasonably close.

clawconduce:
Is the code beautified?  What parameters are used to beautify the source?

bitwiseman:
@clawconduce Thanks for contributing.  Disliking the code is perfectly acceptable and you are in good company.  It's a long road.

If you run, `./build` that will beautify the code, yes.   The settings are in `./jsbeautifyrc`.

js/lib/beautify.js
```diff
@@ -1014,6 +1015,62 @@ function Beautifier(js_source_text, options) {
...
+                        }
+                        continue;
+                    }
+                    if (lookingAt === 'OPERATOR' && check_token.type == 'TK_OPERATOR') {
+                        lookingAt = 'VALUESTART';
+                        continue;
+                    }
+                    if (lookingAt === 'VALUESTART' || lookingAt === 'VALUE') {
+                        if (lookingAt === 'VALUESTART' && setKeyPadding !== null) {
+                            check_token.text = new Array(setKeyPadding - lastKeyLen + 1).join(' ') + check_token.text;
```
clawconduce: this is my main dislike.  I tried setting whitespace_before to a string of whitespace, but that didn't appear to affect the output.  Is there a better way?

bitwiseman: Whitespace_before is a value on the parsed token, but ignored during output except in the case of `keep_array_indentation = true`.  
see https://github.com/beautify-web/js-beautify/blob/master/js/src/javascript/beautifier.js#L413

bitwiseman:
@clawconduce - When do you think you'll have a chance to work on this?

clawconduce:
I probably won't get the chance for a few weeks at least.  If someone else wants to pick this up, I have no problems with that too, though I will hopefully get some time sooner rather than later.

I have been thinking in the back of my mind how to limit when this happens, I realized that when the values are functions, especially long functions, this sort of indentation is not desirable, so either objects with functions get skipped completely, or maybe there's a parameter that will disable this if any value in an object is a function more than N lines long.

bitwiseman:
@clawconduce - Any updates?

clawconduce:
Sorry for the long time between updates.

It seems my todo list is growing quicker than I am pulling stuff off.  I don't myself getting this merge ready anytime soon.

I'll close this PR - hopefully someone sees #396 and can use this as a starting point.

PR #1232, opened 2017-08-17
Added unpack option to js-beautify cli

hnzlmnn:
I added a option (-u, --unpack) to the CLI to automatically run the provided unpackers before starting the beautifying process. The option works for the JavaScript and the Python version. All documentation texts i found are updated too.


bitwiseman:
@hnzlmnn This is an old PR but fixing it to work with the current codebase would address #1350.

README.md
```diff
@@ -127,6 +127,7 @@ CLI Options:
   -v, --version    Show the version
 
 Beautifier Options:
+  -u, --unpack                      Try to unpack before beautifying
```
bitwiseman: `-u` is already used for `unindent-chained-methods`.

bitwiseman:
@hnzlmnn 
Would you be willing to circle back and resolve the merge conflicts here?   Also, even if we can use `-u` this would still be useful.

PR #1241, opened 2017-09-07
Don't publish unnessessary files

simondel:
By ignoring everything at the start of the npmignore and then excluding the files you actually need, you can reduce your node_modules foodprint.

The difference:

| What | Current size | New size | Difference |
|-----:|:--------:|:-----:|:----:|
|node_modules| 2.18 MB | 1.57 MB | 38.85% |
| js-beautify folder | 1.27 MB | 674 KB | 46.93% |
| zipped | 656 KB | 416 KB | 36.58% |

If js-beautify would be downloaded 1.3 milion times in the month after this is released, it would save about 312 GB of traffic due to the smaller zip size.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,4 @@
 **/*
 !js/
-!README.md
-!LICENSE
-!CHANGELOG.md
 !CONTRIBUTING.MD
+# README, LICENSE, CHANGELOG and package.json are always added
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,6 @@
-Makefile
-index.html
-php/
-python/
-web/
+**/*
+!js/
+!README.md
+!LICENSE
+!CHANGELOG.md
+!CONTRIBUTING.MD


simondel:
The `js` folder still contains these files: If any files or folders are not required. Please let me know and I'll update the PR:
image
`<Bestanden>` are just files that are in the root of the folder.

bitwiseman:
@simondel - Are package.json and package-lock.json included?

simondel:
@bitwiseman The package.json is always published. The package.lock.json is *never* published. The [documenation about the npmignore](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) specifies:
>The following paths and files are never ignored, so adding them to .npmignore is pointless:
>* package.json
>* README (and its variants)
>* CHANGELOG (and its variants)
>* LICENSE / LICENCE

I tested this with your readme, changelog and license file see: https://github.com/simondel/tempsimonl892017. This package has been published to npm and does contain the files in the repo. I'll update the PR to no long explicitly include them.

The [package.lock.json documentation](https://docs.npmjs.com/files/package-lock.json) states:
> **One key detail about package-lock.json is that it cannot be published**, and it will be ignored if found in any place other than the toplevel package. It shares a format with npm-shrinkwrap.json, which is essentially the same file, but allows publication. This is not recommended unless deploying a CLI tool or otherwise using the publication process for producing production packages.

bitwiseman:
Great! Thanks!

PR #1249, opened 2017-09-18
Revert "Don't publish unnessessary files."

Kingwl:
This reverts commit 641d3bc6f61726f7302d69005b327582c0123617.


Kingwl:
@bitwiseman @simondel
looks the npmignore ignore all file
so i revert this commit for quick fix

simondel:
#1251 already fixes this

Issue #1187, opened 2017-05-09
Missing space between anonymous async function and first paren


prettydiff:
https://github.com/Glavin001/atom-beautify/issues/1654

After searching the JS Beautify code it looks like there are tests in place for this, but none of them make use of fat arrows.  The primary difference is the test samples all expect the *function* keyword which isn't needed when working with fat arrow.

TwoAbove:
Hey guys! 
Any updates on this?

bitwiseman:
@TwoAbove 
This feature is not assigned an needs someone to write tests and implement it.

TwoAbove:
@bitwiseman I can try. It needs both js and python implementations, right?

bitwiseman:
@TwoAbove - yes.  You can pick which ever language is easier for you and then translate.

bitwiseman:
#896

PR #977, opened 2016-07-22
Add space between async keyword and arrow functions

afuous:
Fixes #896

Issue #896, opened 2016-03-11
Whitespace after ES7 `async` keyword for arrow functions


geakstr:
Hello. js-beautify format

``` js
const double = async (x) => x * 2;
```

to

``` js
const double = async(x) => x * 2;
```

It continues to work, but seems like bad behavior. 

It formatted correctly

``` js
const double = async function(x) { return x * 2; };
const double = async x => x * 2;
```

paullryan:
Given that async/await just made it into node7 it would be great for this to get merged...

bitwiseman:
@paullryan - if you want to pick up for @afuous left off and work PR #977 to completion that would be greatly appreciated.

serge-nikitin:
Any progress? 

eslint-config-airbnb is not happy with the lack of space :(

PR #1264, opened 2017-09-27
Whitespace between async and arrow function

cejast:
Adds whitespace between `async` and an arrow function:

*Input*
`async() => {}`

*Output*
`async () => {}`

Addresses #896 and #1187 and is a continuation of #977

### DIFF
--- a/js/lib/beautify.js
+++ b/js/lib/beautify.js
@@ -878,8 +878,8 @@ function Beautifier(js_source_text, options) {
             }
         }
 
-        // Should be a space between await and an IIFE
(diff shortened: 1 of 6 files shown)


bitwiseman:
@cejast 
This is great!  I can merge this as it is, but would you be willing to add a few more tests?
Most important, a test for the other secnario from #896:
```js
async x => x * 2;
async (x) => { return x * 2; }
```

Basically, whatever parts of the syntax show at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions that you can show works.

This will be great to have fixed, but it would be better to ensure the fix covers the behavior sufficiently that we don't have to turn around and do another change. 

Thanks!

PR #1277, opened 2017-10-17
Made brace_style option more inclusive

Coburn37:
Per the discussion of https://github.com/beautify-web/js-beautify/pull/1058, the `brace_style` option has been made more inclusive. Also 3 shell tests have been added.

* Accepts `expand,collapse,preserve-inline` as `collapse,preserve-inline`
* Accepts `preserve-inline` as `collapse,preserve-inline` (uses the default `brace_style` `collapse`+ user specified `preserve-inline`)
* Accepts `preserve-inline,collapse` as `preserve-inline`
* Rejects `invalid,preserve-line,collapse` as invalid

Whereas in the previous version these would all reject as invalid.


Coburn37:
This also adds an invalidHandler to `nopt` in the javascript `cli.js` which apparently wasn't there before? So invalid parameter wouldn't throw and now they do

PR #1287, opened 2017-11-08
Fix to use correct variable name. js_source_text > source_text

### LINKED ISSUE #1286: Strict mode: js_source_text is not defined [CSS]
Should this line not be `source_text`, rather than `js_source_text`?

https://github.com/beautify-web/js-beautify/blob/6b60339719c1491dce6eda8302f092accbf94ad6/js/src/css/beautifier.js#L199

When including `beautify-css.js` file as a module, which _**always defaults to strict mode**_, the following error will be thrown due to `js_source_text` not being defined.

```shell
Uncaught ReferenceError: js_source_text is not defined
```

This doesn't appear to be used anywhere and I can only assume it's a typo?..

TremayneChrist:
Fixes #1286

### DIFF
--- a/js/lib/beautify-css.js
+++ b/js/lib/beautify-css.js
@@ -335,7 +335,7 @@ function Beautifier(source_text, options) {
             preindent_index += 1;
         }
         baseIndentString = source_text.substring(0, preindent_index);
-        js_source_text = source_text.substring(preindent_index);
+        source_text = source_text.substring(preindent_index);
     }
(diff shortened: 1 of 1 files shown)


bitwiseman:
@TremayneChrist  
Please see CONTRIBUTING.md for how to run build.

bitwiseman:
Thanks for the contribution, I've tweaked it and merged.

Issue #1165, opened 2017-03-24
SCSS Comment Issue


rgfx:
Comments in SCSS are getting sent to line before after update. Expected output? Is there a setting I need to change?

# Input
The code looked like this before beautification:
```
text-align: center;
    // height: auto;
// display: flex;
```

# Expected Output
The code should have looked like this after beautification:
```
text-align: center;
// height: auto;
// display: flex;
```

# Actual Output
The  code actually looked like this after beautification:
```
text-align: center;//height: auto;
// display: flex;
```

## Environment
OS: Windows 7 
Vscode 

## Settings
Everything is default. It didn't do this before.

rgfx:
Found [vscode-stylesheet-beautify](https://marketplace.visualstudio.com/items?itemName=n23.vscode-stylesheet-beautify) for a temp fix.

bitwiseman:
@rgfx - interesting.  That tool is using js-beautify `^1.6.8` internally.

bitwiseman:
Got the following when running on jsbeautifier.org:
Start:
```
<style>
    text-align: center;
//height: auto;
    // display: flex;
</style>
```
First output:
```
<style>
    text-align: center;
    //height: auto;
    // display: flex;
</style>
```
Second output (when beautify called on first output):
```
<style>
    text-align: center; //height: auto;
    // display: flex;
</style>
```

rgfx:
The extension I found before that was using ^1.6.8 updated :( Found this as a replacement for now. https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode you will have to regex your files. 

find: ; //
replace: ; \n //

webberwang:
Getting this issue as well

Issue #1220, opened 2017-07-23
Keep adding new lines between comments in css


zyyedi:
Hi, I just download the plugin for sublime text, and I met some problems. It keeps adding new lines after each reformatting.
<img width="535" alt="screen shot 2017-07-23 at 5 53 54 pm" src="https://user-images.githubusercontent.com/26909419/28498472-de8b67fa-6fd0-11e7-9b85-b28ca549334f.png">
<img width="598" alt="screen shot 2017-07-23 at 5 54 13 pm" src="https://user-images.githubusercontent.com/26909419/28498473-df8e4f46-6fd0-11e7-9646-5f7297a1ce33.png">
<img width="576" alt="screen shot 2017-07-23 at 5 54 22 pm" src="https://user-images.githubusercontent.com/26909419/28498474-e0db5f92-6fd0-11e7-97f9-739315fff55d.png">

bitwiseman:
Duplicate of #1185.

Issue #857, opened 2016-01-29
Weird Beautify Style?


raulfunkie:
it seems that beautify likes to add a double line before a comment, then a single line after a comment.
## Original

``` css
.box {
    border: 2px solid gray;
    padding: 10px;
    background: rgba(0, 0, 0, .5);
}

/* This is a CSS comment! */
.other-box {
    border: 1px solid red;
    padding: 4px;
    background: #fe4444;
}
```
## After beautify

``` css
.box {
    border: 2px solid gray;
    padding: 10px;
    background: rgba(0, 0, 0, .5);
}


/* This is a CSS comment! */

.other-box {
    border: 1px solid red;
    padding: 4px;
    background: #fe4444;
}
```
## Expected Outcome

Same as original

bitwiseman:
Please replace your screenshots with text.

bitwiseman:
Thanks!

raulfunkie:
Is there any solution proposed for this? or is way way below the priority line D:?

bitwiseman:
I haven't had time, sorry.

raulfunkie:
Is this in the road-map for a fix soon? (if ever).

Just wanted to know, thanks!

hainguyents13:
up for this issue, anything new sir @bitwiseman ?

PR #1291, opened 2017-11-19
Ensure comments maintain position

amandabot:
Block comments above rule sections will no longer add a newline below. #531 
Single line comments below a property will not jump to the same line as the property. #1165 

This request covers part of #1145 (which seems to have been in limbo for several months) , but leaves out the CLI option. 

### Input

```js
/* block comment */
.tabs { 
    width: 10px;
    // comment follows rule
    // another comment new line
}
```

### Output

```js
/* block comment */
.tabs { 
    width: 10px;
    // comment follows rule
    // another comment new line
}
```

One caveat: I had trouble implementing a working test for #531. When I run a similar input to what is  above in a file locally through the CLI, the output is correct. When I setup a similar test in the tests, the output is different. Perhaps, in my newness to the code base, I am missing something?

Test I tried:
```js
{
    input: '.tabs{width: 10px;\n// comment follows rule\n// another comment new line\n}',
    output: '.tabs {\n\twidth: 10px;\n\t// comment follows rule\n\t// another comment new line\n}'
}
```

I get this as output:
```js
.tabs {
        width: 10px; // comment follows rule
        // another comment new line
}
```


bitwiseman:
@amandabot
Sorry for slow response, I was  on vacation. 
The issue you're seeing with testing #531 is very likely a setting difference. 
Maybe `preserve_newlines`?

bitwiseman:
@amandabot  Don't Worry about the extra commits.  I can squash them when I merge the PR.  
The style issues you encountered are from code climate which doesn't run locally.  Nothing to be done about it.

test/data/css/tests.js
```diff
@@ -304,7 +304,7 @@ exports.test_data = {
                 { unchanged: '/* test */' },
                 { input: '.tabs{/* test */}', output: '.tabs {\n\t/* test */\n}' },
                 { input: '.tabs{/* test */}', output: '.tabs {\n\t/* test */\n}' },
-                { input: '/* header */.tabs {}', output: '/* header */\n\n.tabs {}' },
+                { input: '/* header */.tabs {}', output: '/* header */\n.tabs {}' },
```
bitwiseman: It doesn't seem like this behavior should change. 
I think the header followed by a blank line should be maintained.

test/data/css/tests.js
```diff
@@ -322,9 +322,49 @@ exports.test_data = {
                 { input: '.tabs{//comment\n//2nd single line comment\nwidth:10px;}', output: '.tabs {\n\t//comment\n\t//2nd single line comment\n\twidth: 10px;\n}' },
                 { input: '.tabs{width:10px;//end of line comment\n}', output: '.tabs {\n\twidth: 10px; //end of line comment\n}' },
                 { input: '.tabs{width:10px;//end of line comment\nheight:10px;}', output: '.tabs {\n\twidth: 10px; //end of line comment\n\theight: 10px;\n}' },
-                { input: '.tabs{width:10px;//end of line comment\nheight:10px;//another\n}', output: '.tabs {\n\twidth: 10px; //end of line comment\n\theight: 10px; //another\n}' }
+                { input: '.tabs{width:10px;//end of line comment\nheight:10px;//another\n}', output: '.tabs {\n\twidth: 10px; //end of line comment\n\theight: 10px; //another\n}' },
```
bitwiseman: I'd also like to keep this test the same if possible, or at least show it with the same input. 
You have another copy of this  new input below at line 362, so need to change it here.

amandabot: Would you mind clarifying what you're asking here? 
325 hasn't changed (except the comma at the end because I added another rule after), and the tests on 326 and 362 demonstrates that comments on the line following a rule are not affected by formatting.

test/data/css/tests.js
amandabot: I would like to present the case that this behavior *should* change. 

Zero-level CSS block comments are formatted differently from those on other levels. This seems unintuitive to me. In the absence of a specific directive to make changes, a beautifier should leave a code unchanged. 

I suspect, with the exception of horizontal indentation, people generally place comments where they intend them to be with respect to vertical space. 

There have been multiple issues opened on comment placement, most specifically about the types of comments this pull request addresses:

Extra newline is inserted after the comment line instead of before it #531
Keep adding new lines between comments in css #1220 
css-beautify inserts extra newlines before comments #1185
css beautify adding an extra newline after a comment line in a css block #609
Weird Beautify Style? #857
How to keep comments on their own lines after formating #1348
SCSS Comment Issue #1165

A recurring theme is that they expect comments to stay where they put them.

On the opposing side, I found an issue requesting the current behavior of the beautifier here:

Blank line before and after CSS / JS comments #736

There are several paths we could take:
* Change the current behavior
* Maintain the current behavior
* Add a CLI/config option for either way.
* Open this as a discussion issue and ask the community

Changing the behavior should not affect existing code as my pull request preserves the vertical locations of comments (while still respecting newline preservation options).

The second option leaves several issues unresolved.

I would argue a CLI option is overly-specific, and it risks opening the door to targeting more specific use cases rather than broadly applicable rules. If you feel this is the way to go, I can reuse some of the work from here #1145 since the pull request has been stagnant for some time.

Let me know what you you think.

bitwiseman: @amandabot 
Whoa.  Okay, let me look at this a bit more.

bitwiseman:
@amandabot 
I updated the PR with a matrix based test and a fix to the python implementation that wasn't preserving more than one newline after inline-comments. 

This change fixes a lot of issues, but interestingly this still doesn't address #531.  That has to do with `newline_between_rules` adding a blank line when a rule end is detected, but not detecting that there are comments before the next rule.  That is no reason not to accept this PR though.  😄

bitwiseman:
Hm. #531 and #857 have the same problem here.

bitwiseman:
@amandabot
Thanks for your great contribution and seeing it through over such a long period.  

This change is major enough that I'm probably going to have to bump the next release to 1.8.x. 

Thanks again!

PR #1292, opened 2017-11-19
Fixed and updated JavascriptObfuscator

mixalbl4-127:
1) fixed bug with replacing value like: "'123 $'" (very big bag)
2) added double quote support in var _0x...=["..."]
3) added function replace support instead _0x... variable (ex:
```
  var _0x311a=["\x74"];
  var _0xa311 = function(z) {
      return _0xa311[z];
  };
  alert(_0xa311(0));
```
)


bitwiseman:
@mixalbl4-127 Please run `./build` and commit any changed files. 
Also, we'll need tests demonstrating the fixes.

bitwiseman:
@mixalbl4-127  Closing due to no response.  Will reopen if updated.

Issue #1262, opened 2017-09-27
Wrap align html attributes when they rich certain amount of characters


asolopovas:
Is it possible to only wrap html attributes when line reaches certain length of characters. The way it works in Phpstorm for example is that If I have something like this.

```html
<div v-for="item in items" class="items">
    <p>{{item.name}}</p>
</div>
```
Phpstorm formatter does not wrap the attributes but if my code looks something like this 
```html
<div v-for="item in items" class="items" v-if="items.notEmpty()" :class="{active: isActive} :key="item.id">
    <p>{{item.name}}</p>
</div>
```
and line contains more then 100 characters then it will wrap the attributes and align them like so

```html
<div v-for="item in items" 
     class="items" 
     v-if="items.notEmpty()" 
     :class="{active: isActive} 
     :key="item.id">
    <p>{{item.name}}</p>
</div>
```

bitwiseman:
See #1125 for related discussion.  Not exactly the same feature, but related.

RenaldasK:
Is this getting implemented? It's very annoying when the following:
`<meta content="IE=edge" http-equiv="X-UA-Compatible">`

becomes:
```
<meta content="IE=edge"
      http-equiv="X-UA-Compatible">
```

Even though there's plenty of space to keep it in one line

hermbit:
+1, this is exactly how we would expect the behavior to work

RenaldasK:
I might take a look it this weekend. Should this be implemented under `--wrap-attributes auto` or some new option?

Adondriel:
I was thinking an option for something like a "number of attributes before wrapping" would be nice, so I could say, anything with < 4 attributes won't be wrapped, while anything >= 4 attributes would be wrapped, and that would be configurable per user, of course.

vvs:
@Adondriel This is tricky. One attribute could be just 3-5 characters long, another one can be 50 chars. Wrapping by default should deal with lengths, not with number of attributes, but it is an interesting additional option to consider...

Issue #1285, opened 2017-11-02
Wrap and align html attributes when line reaches wrap-line-length


Helayxa:
# Description
I would like to get the combination of the power of :
`"wrap-attributes": "force-aligned",`
`"wrap-line-length": 120`
If my line reaches the wrap-line-length, it should wrap the whole line and align them with the previous line.

# Input
The code looked like this before beautification:
```html
<p-dataTable
    [value]="data"
    [(selection)]="selectedRows"
    (selectionChange)="onSelectionChange($event)"
    [editable]="editable"
    [selectionMode]="settings.selectionMode"
    [resizableColumns]="settings.resizable"
    [reorderableColumns]="settings.reorderable"
    responsive="true"
    scrollable="true"
    scrollHeight="550px"
    [lazy]="settings.lazy"
    [totalRecords]="totalRecords"
    [paginator]="settings.paginator"
    [rows]="settings.paginator?.rowsPerPage || 10">
</p-dataTable>
```

# Expected Output
The code should look like this after beautification with that feature:

```html
<p-dataTable [value]="data" [(selection)]="selectedRows" (selectionChange)="onSelectionChange($event)" [editable]="editable"
             [selectionMode]="settings.selectionMode" [resizableColumns]="settings.resizable" [reorderableColumns]="settings.reorderable"
             responsive="true" scrollable="true" scrollHeight="550px" [lazy]="settings.lazy" [totalRecords]="totalRecords"
             [paginator]="settings.paginator" [rows]="settings.paginator?.rowsPerPage || 10">
</p-dataTable>
```

cheerypick:
+1, me and my team need the same feature, and it seems to be the same feature as requested here: https://github.com/beautify-web/js-beautify/issues/1262
I saw this exact feature requested at VSCode github too: https://github.com/Microsoft/vscode/issues/2204 (last comment)

I played with the source code a bit and have some implementation that seems to work: a new option for "wrap_attributes" - "aligned" which doesn't force the attrs to wrap unless the line length is reached, and aligns them properly when the line is wrapped.

Can something like this work, @bitwiseman? 

If so, we could discuss the details or I can open a PR. (But to be honest, I am not exactly an open source expert and a bit confused about the tests and python version and whatnot). Well, html beautifier doesn't exist in python so maybe it's not an issue :) So guidance is appreciated!

bitwiseman:
@cheerypick 
Thanks for taking a swing at implementing this.  

[...]

PR #1297, opened 2017-11-23
Add 'aligned' option that align attributes after specific line length…

cheerypick:
This PR is about requested soft-alignment feature of attributes alignment after and only after specific line length threshold is reached. 

related to https://github.com/beautify-web/js-beautify/issues/1262

Basically, this PR adds a new option that doesn't match 'is_wrap_attributes_force' condition (that forces line break after each attribute) but indents the attributes same way as  'is_wrap_attributes_force_aligned' if 'wrapped' is set to true by wrap_line_length. 

//At least, I suppose it does :)

As I said previously, I am not exactly an open source expert, so comments, guidance, and advice appreciated. 
(Hope that I understood how the test matrices work correctly too)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ as well as deobfuscate scripts processed by
 # Contributors Needed
 I'm putting this front and center above because existing owners have very limited time to work on this project currently.
 This is a popular project and widely used but it desperately needs contributors who have time to commit to fixing both
-customer facing bugs and underlying problems with the internal design and implementation.  
+customer facing bugs and underlying problems with the internal design and implementation.
 
 If you are interested, please take a look at the [CONTRIBUTING.md](https://github.com/beautify-web/js-beautify/blob/master/CONTRIBUTING.md) then fix an issue marked with the ["Good first issue"](https://github.com/beautify-web/js-beautify/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) label and submit a PR. Repeat as often as possible.  Thanks!
 
@@ -94,7 +94,7 @@ You can also use `js-beautify` as a `node` library (install locally, the `npm` d
 $ npm install js-beautify
 ```
 
-Import and call the appropriate beautifier method for javascript (js), css, or html.  All three method signatures are `beautify(code, options)`. `code` is the string of code to be beautified. options is an object with the settings you would like used to beautify the code.  
+Import and call the appropriate beautifier method for javascript (js), css, or html.  All three method signatures are `beautify(code, options)`. `code` is the string of code to be beautified. options is an object with the settings you would like used to beautify the code.
 
 The configuration option names are the same as the CLI names but with underscores instead of dashes.  For example, `--indent-size 2 --space-in-empty-paren` would be `{ indent_size: 2, space_in_empty_paren: true }`.
 
@@ -210,7 +210,7 @@ Configuration sources provided earlier in this stack will override later ones.
 
 The settings are a shallow tree whose values are inherited for all languages, but
 can be overridden.  This works for settings passed directly to the API in either implementation.
-In the Javascript implementation, settings loaded from a config file, such as .jsbeautifyrc, can also use inheritance/overriding.  
+In the Javascript implementation, settings loaded from a config file, such as .jsbeautifyrc, can also use inheritance/overriding.
 
 Below is an example configuration tree showing all the supported locations
 for language override nodes.  We'll use `indent_size` to discuss how this configuration would behave, but any number of settings can be inherited or overridden:
@@ -239,7 +239,7 @@ for language override nodes.  We'll use `indent_size` to discuss how this config
 Using the above example would have the following result:
 
 * HTML files
-  * Inherit `indent_size` of 4 spaces from the top-level setting.  
+  * Inherit `indent_size` of 4 spaces from the top-level setting.
   * The files would also end with a newline.
   * JavaScript and CSS inside HTML
     * Inherit the HTML `end_with_newline` setting.
@@ -289,8 +289,8 @@ HTML Beautifier Options:
   -b, --brace-style                  [collapse-preserve-inline|collapse|expand|end-expand|none] ["collapse"]
   -S, --indent-scripts               [keep|separate|normal] ["normal"]
   -w, --wrap-line-length             Maximum characters per line (0 disables) [250]
-  -A, --wrap-attributes              Wrap attributes to new lines [auto|force|force-aligned|force-expand-multiline] ["auto"]
-  -i, --wrap-attributes-indent-size  Indent wrapped attributes to after N characters [indent-size] (ignored if wrap-attributes is "force-aligned")
+  -A, --wrap-attributes              Wrap attributes to new lines [auto|force|force-aligned|force-expand-multiline|aligned-multiple] ["auto"]
+  -i, --wrap-attributes-indent-size  Indent wrapped attributes to after N characters [indent-size] (ignored if wrap-attributes is "aligned")
   -d, --inline                       List of tags to be considered inline tags
   -U, --unformatted                  List of tags (defaults to inline) that should not be reformatted
   -T, --content_unformatted          List of tags (defaults to pre) whose content should not be reformatted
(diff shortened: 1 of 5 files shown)


bitwiseman:
Thanks the PR.  You understood exactly how the test matrices work.  
Because of that I can say that your change doesn't quite address #1262.  

In #1262, they ask for: 
```html
<div v-for="item in items" 
     class="items" 
     v-if="items.notEmpty()" 
     :class="{active: isActive} 
     :key="item.id">
    <p>{{item.name}}</p>
</div>
```

Whereas this PR would allow for something like (depending on the wrap line length): 
```html
<div v-for="item in items" class="items" v-if="items.notEmpty()" 
     :class="{active: isActive} :key="item.id">
    <p>{{item.name}}</p>
</div>
```

To do #1262, you'll need to look-ahead to see if the current element with attributes is going to wrap and then switch to force-align mode if that is the case.  

In terms of quality, this is a fine PR, it just needs some to be changed to implement the feature described in #1262.

cheerypick:
Hmm,  good point, I think it can be discussed. I was in first place solving my own problem and the issue in https://github.com/beautify-web/js-beautify/issues/1285, and didn't quite address what was requested in https://github.com/beautify-web/js-beautify/issues/1262. 

Frankly speaking, what I was trying to recreate was Intellij's behavior: it doesn't wrap attributes before certain length but wraps and aligns them every time the line is over a threshold. If there is room for more than one attribute, so it tries to fit them, otherwise wraps. Our code style (grown from the time when the majority of devs on the project were using IntelliJ) doesn't impose that every attribute is on the new line, otherwise, we could just use force-align.  (And personally, I think it's a bit strange behavior - like, we are fine with 3-5 attributes until line is below threshold, but suddenly if the threshold is reached, so it's one per line). I thought that all JetBrains' products behave like this, but I maybe not, and I totally understand that people may want such behavior :) I am not familiar how code formatting is configured in these IDEs and why it is different, though.

Furthermore, what was requested in https://github.com/beautify-web/js-beautify/issues/1285, demonstrates kinda the opposite behaviour - it removes existing line breaks and reindents everything according to line length and indents. I am not sure how this should work with preserve-newlines? Not sure either that my implementation will handle this correctly. 

So in order to make more people happy (at least me, author of 1262 and 1285), we should maybe provide some combination of options that enables all the scenarios? Like, 'soft' indented wrapping, and force wrapping but after certain initial line length. Any architectural tips on that?

bitwiseman:
@cheerypick
I've reopened #1285 since you've demonstrated that it is different from #1262.  
I'm happy to have you address #1285 now and leave #1262 to later (or someone else).  Some combination of options is exactly what will be needed.  What set of setting names would you use to describe these?  (Ignore the current naming if you like, I'm open to changing the values if it will make them consistent and  easier to understand.)

bitwiseman:
@cheerypick  Any chance to look at this further?

cheerypick:
Oh, @bitwiseman, sure. Had to prioritize other things, but will look at it hopefully this week.

bitwiseman:
@cheerypick Bump.  I understand if you're busy, just checking.

RenaldasK:
Not really sure what is holding this up, but I might fork this and try to get it merged and closed, as I would love to see this feature in the next release.

bitwiseman:
@RenaldasK 
Thanks for the question.  I had to go back and read again to make sure.  

@cheerypick  said: 

> So in order to make more people happy (at least me, author of 1262 and 1285), we should maybe provide some combination of options that enables all the scenarios? Like, 'soft' indented wrapping, and force wrapping but after certain initial line length. Any architectural tips on that?

I said I was fine if we wanted to have two separate PRs - this one to address #1285, another to address #1262.  

If we go that route I'd still like to have the naming figured out - If #1285 behavior is `aligned`, what is #1262 going to be called? I want to have some plan for that.  This naming structure for attribute wrapping is reaching its limits and will likely need to be designed soon anyway. Thus I asked @cheerypick what she thought the names should be. 

@cheerypick also, mentioned a number of other scenarios that might or might not have side-effects in this area.  I wasn't sure how to proceed.  

Separately, looking at the complexity of this area, there probably needs to be another test or two, or even another matrix entry - maybe a `line_wrap=40` case.  

Once again, this is solid PR as it stands, I was mostly waiting on @cheerypick to make the call about what direction she wanted to go - Adding #1262 or just doing #1285.

michaelauderer:
For the naming, here's what I propose:
#1285 is `soft-aligned`
#1262 is `aligned`

#1285 is "soft" because it doesn't align every single attribute, just the lines themselves. 
#1262 makes more sense to be called "aligned" imo because it's truly aligning the attributes- its the same behavior as force-aligned without being forced.

vvs:
Hey @bitwiseman, @cheerypick, @michaelauderer, other folks that are interested.

Here's my take on the topic.

## What we have 

`wrap_attributes`:

* `auto` - enables wrapping of attributes.
* `force` - first attribute stays, others wrapped with "wrap_attributes_indent_size"
* `force-aligned` - the first attribute stays, others wrapped and aligned with the first one
* `force-expand-multiline` - If there are more than 1 atribute, *ALL* attributes are wrapped no matter what and indented with "wrap_attributes_indent_size". AND: the closing '>' of the opening tag is also on the new line. This mode is currently broken, produces wrongly indented files :(

I think that there are multiple (orthogonal) concerns here and putting them all into single value would create more confusion and weird option names explosion if we are to handle all cases.

So, let's separate wrapping, aligning and number of attributes per line, etc.

## Proposal

`wrap_attributes_mode`:

* `none` - We just don't wrap attributes, period. (needed?)
* `auto` - Normal attribute wrapping, this should be the default, probably, as the most logical case. When the line is too big, we wrap. If all attributes can fit into a single line, they stay on the single line, no wrapping. (Default.)
* `force` - We forcefully wrap if there are more than 1 attribute present, otherwise auto.

When the wrapping is in action, the following properties are in play. If no wrapping happens, they don't play any role.

`wrapped_attributes_per_line`:

* `multiple` - Soft wrapping, *default*. Fit as many attrs per line as allowed.
* `single` - The first attribute stays on the first line, others are on separate lines.
* `single-all` - All attributes, including the first one, placed on separate lines. So, every attribute is on a separate line, and the tag line has no attributes.

`wrapped_attributes_indent`:

1. auto -- Uses `wrap_attributes_indent_size`, *default*.
2. aligned -- All attributes start aligned with the first one. If the first attr is on a new line ("single-all" mode of `wrapped_attributes_per_line`), then we use `wrap_attributes_indent_size` to calculate the indent.

`wrapped_attributes_end`:

* `auto` - *default*
* `multiline` - the closing `>` is on the separate line.

## Examples 

1. DEFAULT case: We soft wrap and try to fit as many attributes per line as possible, but within the line length limit.
```
wrap_attributes_mode = auto
wrapped_attributes_per_line = multiple
wrapped_attributes_indent = auto
wrapped_attributes_end = auto
```

2. #1285 case. Very similar to default case, but we align the wrapped attributes to be on the same indent level as the first attribute.
```
wrap_attributes_mode = auto
wrapped_attributes_per_line = multiple
wrapped_attributes_indent = aligned
wrapped_attributes_end = auto
```
3. #1262 case. When we wrap, we want one attribute per line, aligned with the first one.
```
wrap_attributes_mode = auto
wrapped_attributes_per_line = single
wrapped_attributes_indent = aligned
wrapped_attributes_end = auto
```
4. Current `force-expand-multiline` case: Force wrapping if there are more than 1 attribute, with one attribute per line, and with multi-line ending, placing the closing `>` on the new line. 
```
wrap_attributes_mode = force
wrapped_attributes_per_line = single-all
wrapped_attributes_indent = auto
wrapped_attributes_end = multiline
```

bitwiseman:
@vvs 
Wow, thanks! That is a great proposal, really well thought out and covers all the current combinations and then some.  That said, I'd like to make this PR a non-breaking change.
  
Could you open a new issue and put your whole comment in there? 

@michaelauderer 
To match approximately to @vvs's naming proposal #1285 (this PR) would use `aligned-multiple` and #1262 (a future PR) would use `aligned-single`.   If that sounds good to you, please make the change and submit a PR to amend this PR.   

Oh, one other thing we need in order to accept this change - update the README.md with the new option.  

Thanks everyone!

michaelauderer:
@bitwiseman Unfortunately I don't, but it looks like you have it working now!

I see that this is part of v1.7.6 milestone, but there are a few other issues blocking that release.

However, since this has been a highly requested feature for a long time (we really need this fix pulled into VSCode), would it be possible to release this now and move the remaining 4 issues to v1.7.7?

PR #1303, opened 2017-12-13
Add support for BigInt literals in JavaScript

thejoshwolfe:
BigInt is an upcoming feature in JavaScript. Read more here: https://github.com/tc39/proposal-bigint

In writing BigInt tests for test262, I found js-beautify useful for formatting my generated tests, so I added bigint literal parsing support here for my own uses. Here is a PR of what I did, which I believe should be of general interest.

The JS and Python tokenizers were similar enough that it was an extra mere seconds to port this change to the other implementation. Great work keeping that easy!

I did not include any tests in this PR, but here is the test case I used to hit every code path I added/edited:

```js
const a = 1n;
const b = .0n;
const c = 1.0n;
const d = 1e0n;
```

I looked into adding this to the test framework here, but it was just a little bit too hard to do right away. My primary task at the moment is to write test262 tests, not maintain js-beautify. But thought I'd share what I had nonetheless.


bitwiseman:
@thejoshwolfe 
Thanks for contributing.  I'm glad you found the beautifier useful.

I'm not opposed to adding this - it seems like a non-breaking change.  However, to take changes I'll need you to make changes to files under `src` (not under `lib`), add tests to the files under `test`, and run the build successfully.   

I understand this is not your primary task, but that's the bar we have to set to maintain the stability and usefulness of this tool.

thejoshwolfe:
> However, to take changes I'll need you to add tests to the files under test, and run the build successfully.

I looked at `test/data/javascript/tests.js`, and I was overwhelmed. But I suppose we could discuss what to do in this thread.

With the example I pasted above, thoroughly testing this feature would sometimes result in invalid JavaScript. Three of the following lines are SyntaxErrors:

```js
const a = 1n;   // valid
const b = .0n;  // SyntaxError
const c = 1.0n; // SyntaxError
const d = 1e0n; // SyntaxError
```

Is there a mechanism or precedent for testing inputs that are invalid? From the behavior of this program, it looks like detecting SyntaxErrors is out of scope, and the code simply gets rendered like this:

```js
const a = 1n;
const b = .0 n;
const c = 1.0 n;
const d = 1e0 n;
```

Should I put that in the test as the "expected" behavior? Or perhaps js-beautify's behavior is only well defined for inputs that do not contain SyntaxErrors; if that's the case, we should not test invalid cases at all, and it would actually make the implementation here much simpler, because you could just always check for the `n` suffix instead of disabling the check in some cases (eliminating the `allow_bigint` variable).

bitwiseman:
@thejoshwolfe 
Yes, test for the invalid output.  See this section: 
https://github.com/beautify-web/js-beautify/blob/master/test/data/javascript/tests.js#L2696-L2778

The beautifier is a forgiving and only partial parser, but it does not throw.  Instead tries to recover and continue beautifying.  So, encoding the behavior of invalid syntax is still useful but so is keeping the `allow_bigint`.    Thinking about it, it would probably make more sense to have it keep these same input if it encountered strange numbers, but that is a separate task.

bitwiseman:
@thejoshwolfe 
Any chance you can create these tests?  I'd like to release the next version soon and it would be great to get this feature added.

thejoshwolfe:
Sorry, I have no time to work on this anymore. Feel free to close this PR or someone else take over.

Issue #987, opened 2016-08-07
Error in --eol processing in python


bitwiseman:
See commit below for test.

PR #1315, opened 2018-01-02
js-beautify: fix handling for --eol and --outfile

thoger:
Missing comma in the array of long command line options passed to getopt
causes options --eol and --outfile to either be handled incorrectly or
not recognized at all.

### DIFF
--- a/python/jsbeautifier/__init__.py
+++ b/python/jsbeautifier/__init__.py
@@ -210,7 +210,7 @@ def main():
 
     try:
         opts, args = getopt.getopt(argv, "s:c:e:o:rdEPjabkil:xhtfvXnCO:w:",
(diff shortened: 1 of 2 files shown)


bitwiseman:
@thoger 
Thanks!

I think this might close #987, could you try uncommenting the shell test from the python tests in https://github.com/bitwiseman/js-beautify/commit/1f5285d6a47b249db410d10bb916195af679d75c#diff-c93d56d051ea4d3db39c4fc0a1737aed and see if it succeeds after your change?

thoger:
I uncommented the test and the shell-smoke-test.sh passed for me locally.  Hence I updated this PR with additional commit enabling the test.

PR #1316, opened 2018-01-02
js-beautify: fix usage information

thoger:
- All command line options use dashes as separators and not underscores.
  Hence fix --space_after_anon_function and --end_with_newline in
  usage().
- Add --comma-first and --operator-position to usage information.
- Fix few whitespace inconsistencies.
- Do not list <outfile> as standalone argument.  It has to be specified
  as argument for the --outfile command line option.
- Properly import sanitizeOperatorPosition to validate
  --operator-position option argument.
- Re-word beautify_file() exception text, as only one file can be
  specified as command argument.

### DIFF
--- a/python/jsbeautifier/__init__.py
+++ b/python/jsbeautifier/__init__.py
@@ -117,7 +117,7 @@ def beautify_file(file_name, opts = default_options() ):
             stream = sys.stdin
             input_string = ''.join(stream.readlines())
         except Exception as ex:
-            print("Must pipe input or define input file.\n", file=sys.stderr)
+            print("Must pipe input or define at least one file.\n", file=sys.stderr)
             usage(sys.stderr)
             raise Exception()
     else:
--- a/python/jsbeautifier/__init__.py
+++ b/python/jsbeautifier/__init__.py
@@ -9,7 +9,7 @@ import errno
 import copy
 from jsbeautifier.__version__ import __version__
 from jsbeautifier.javascript.options import BeautifierOptions
-from jsbeautifier.javascript.beautifier import Beautifier
+from jsbeautifier.javascript.beautifier import Beautifier, sanitizeOperatorPosition
 
 #
 # The MIT License (MIT)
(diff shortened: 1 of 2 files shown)


python/jsbeautifier/__init__.py
```diff
@@ -117,7 +117,7 @@ def beautify_file(file_name, opts = default_options() ):
             stream = sys.stdin
             input_string = ''.join(stream.readlines())
         except Exception as ex:
-            print("Must pipe input or define at least one file.", file=sys.stderr)
+            print("Must pipe input or define input file.\n", file=sys.stderr)
```
bitwiseman: This text should be identical between the python and javascript implementations.  
See test https://github.com/bitwiseman/js-beautify/blob/master/js/test/shell-smoke-test.sh#L16 
How about the making the text be "Must pipe input or define at least one input file.\n" for both implementations?

python/jsbeautifier/__init__.py
```diff
@@ -9,7 +9,7 @@
 import copy
 from jsbeautifier.__version__ import __version__
 from jsbeautifier.javascript.options import BeautifierOptions
-from jsbeautifier.javascript.beautifier import Beautifier
+from jsbeautifier.javascript.beautifier import Beautifier, sanitizeOperatorPosition
```
bitwiseman: Doh.  Another case of needing more cli option tests.

python/jsbeautifier/__init__.py
thoger: My concern was that the wording of "at least one file" indicates you can specify one or *more* files as argument for the tool.  However, the python CLI only accepts one.  Apparently, that's difference from the js CLI.

Anyway, I'm ok to revert this change if consistency with the js version is preferred, and also update PR #1317 to also use "at least one file".

bitwiseman: @thoger 
Or you could get the python version to take more than one file.  That would be awesome. 😄

thoger: I reverted the wording change to unblock this PR.  A change to handle more than one input file is out of scope of this PR, IMO.

PR #1317, opened 2018-01-02
css-beautify: python cssbeautifier cli

thoger:
This adds a command line tool for the python cssbeautifier.  It is based
on and re-uses as much code as possible from the js-beautify tool.

Long command line options make it possible to control all currently
implemented options in the CSS Beautifier.  Short command line options
are only supported for options that are identical to those of the
js-beautify tool.

Note that there is an inconsistency between the two tools caused by the
different default for the preserve_newlines option - True for
jsbeautifier and False for cssbeautifier, hence different command line
options --disable-preserve-newlines vs. --preserve-newlines.

### DIFF
--- a/python/css-beautify
+++ b/python/css-beautify
@@ -1,0 +1,8 @@
+#! /usr/bin/env python
+#
+# Stub script to run cssbeautifier
+#
+import sys
+from cssbeautifier import main
+sys.exit(main())
+
--- a/python/cssbeautifier/__init__.py
+++ b/python/cssbeautifier/__init__.py
@@ -25,9 +25,13 @@
 
 from __future__ import print_function
 import sys
+import os
+import io
 import re
 import copy
+import getopt
 from jsbeautifier.__version__ import __version__
+from jsbeautifier import isFileDifferent, mkdir_p
 from cssbeautifier.css.options import BeautifierOptions
 from cssbeautifier.css.beautifier import Beautifier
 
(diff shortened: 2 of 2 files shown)


thoger:
As noted in the commit message, I intentionally avoided defining short command line options for CSS Beautifier specific options, leaving them to project maintainer to pick whatever options they prefer.

It might also be worth considering the use of consistent options naming scheme for boolean options in both python CLI tools, that would additionally make it possible to change defaults.  I.e. for every boolean option foo, recognize both --foo and --disable-foo (or --no-foo) regardless of the default.  However, that would make handling of short command line options more difficult.

bitwiseman:
@thoger 
Very cool! 

Would it be possible to have both cli files call into a common helper (see the javascript implementation lib/cli.js). Then there would be less duplication of code and the to implementations would be that much closer to parity.

thoger:
I assume it would be possible.  Whether it's worth the effort I am not sure.  I don't really like the idea of having js-beautify accept and silently ignore options specific to css-beautify and vice versa.  AFAICS, that's the behaviour of the JS CLI (and not only because of quirks of the nopt module).

I do understand the desire to keep JS and PY CLI options consistent, but there are already inconsistencies.  One concern regarding boolean flags and different defaults was raised above.  However, as the nopt used by JS CLI automatically handles --no-foo, that's what PY CLI should do as well.  Should the two versions also be bug-by-bug compatibles?  The nopt does few really odd things:

* If option is configured as to required an argument of certain type, and no argument is provide and the option is the last thing on the command line, it is handled as boolean flag, bypassing validation.  So `js-beautify.js --indent-size` is effectively equivalent to `js-beatufiy.js --indent-size 1`.

* Unknown options are silently accepted.  Rather bad if you make a typo is some long option.

bitwiseman:
"Should the two versions also be bug-by-bug compatibles" - that is, of course, not an objective. :)  However, though there are inconsistencies, in terms of features and behavior the project would like to continue to reduce those differences over time.  The way that has been generally done but maintaining feature parity where we have it and requiring it when fixing bugs or adding new features.  

I'm not trying to punish you for fixing issues (thanks again, it is great to have help).  
I'll create an issue with notes from this discussion so that it is tracked somewhere for the next time someone touches this area.

Issue #1304, opened 2017-12-20
HTML - Line breaks changing code output


valtism:
# Description
I have this problem with the html formatter in VSCode, which I believe is done with js-beautify.

In this case I have a label and an input. I don't want any space between them so I can style them as a single box. 

When there is whitespace or a line break between the elements, spacing is automatically added between the elements.

When I put the elements together in this way, the formatter automatically breaks them up onto separate lines, changing the output. I don't think that the formatter should be changing the meaning or output of the code it is formatting.

# Input
The code looked like this before beautification:
```
<label>Every</label><input class="scheduler__minutes-input" type="text">
```

# Expected Output
The code should have looked like this after beautification:
```
<label>Every</label><input class="scheduler__minutes-input" type="text">
```

# Actual Output
The  code actually looked like this after beautification:
```
<label>Every</label>
<input class="scheduler__minutes-input" type="text">
```

## Environment
OS: Windows 10
Editor: VSCode

## Settings
Default VSCode settings.

bitwiseman:
Yes, this is an issue with the way the beautifier understands HTML.  It is a bug and it would be great to get it fixed.  

The short term workaround would be to wrap those two elements in a `<span>`.  Then they'd stay together.

andreyvolokitin:
But beautifier actually needs to break such tags in case of minified HTML — it consists of them. Can these use cases be distinguished and handled separately? At first look, it seems like pretty hard deal

andreyvolokitin:
[This line](https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify-html.js#L1037) makes this, so it is always a new line at a tag start, if this tag gets formatted. I think this is the place for some additional logic, but not sure how this logic should behave...

andreyvolokitin:
This probably should be an option, because in case of aggressive minification all whitespace between tags gets removed, and then there is no way to distinguish where this whitespace should remain removed after beautification. With an option like ```keepCollapsedWhitespace: true``` beautifier would not add a new line at the start of a tag which has a direct left sibling tag and has no left whitespace. Then the user can beautify minified HTML with this option turned off, and even with this option turned on if HTML will be minified in "conservative" way with 1 space left between tags

bitwiseman:
There's a bunch of discussion around this in #840, #841.

andreyvolokitin:
[...]

PR #1329, opened 2018-01-21
HTML: Add option to not wrap collapsed tags, and bugfix

andreyvolokitin:
See https://github.com/beautify-web/js-beautify/issues/1304

```keep_collapsed_whitespace:  true``` will keep any tag (```TK_TAG_START``` or ```TK_TAG_SINGLE``` tokens) from wrapping on a new line if it has another tag (```TK_TAG_END``` or ```TK_TAG_SINGLE``` tokens) as a direct previous sibling (that is, if ```TK_CONTENT``` between these tokens is empty). For the simplicity, in any other case/token sequence, this option will not take effect, including the following example:
```html
<tag></tag>&nbsp;<tag></tag>
```
That is, this option won't take effect if ```TK_CONTENT``` is a string without whitespace. I am not sure if including this feature may be helpful or not, either way, it can be added later if required. Also, I did not take into account any handlebars content/tokens completely, though probably it would worth investigating, I think it may overcomplicate things...

I also made 2 bugfixes:
- ```record_tag()```:  ```tag``` should be trimmed, because sometimes it can contain whitespace (like ```tag\n```)
- ```print_newline()```:  ```line_char_count``` should be reset to 0 only if wrapping is actually happening


bitwiseman:
@andreyvolokitin - Take a look at the latest release.

andreyvolokitin:
I ran tests from this PR on the latest release with `unformatted: []` option, and also with no options, and these tests have failed

bitwiseman:
Inline and the new html tokenizer superceeds this.  Thanks for contributing.  I hope you will do so again in the future.

PR #1340, opened 2018-02-13
Prevent indent after PHP echo tag (<?= ... ?>)

### LINKED ISSUE #1339: Beautification of PHP with echo short tags
# Description
The formatting in PHP files including HTML is incorrect when the PHP echo short tag is used. I can solve this in my project with this option:

```
void_elements: ['?=']
```

# Input
```
<div class="modal-body">

    <?= $form->field($model, 'Categoria') ?>
    
    <?= $form->field($model,'MontoMin') ?>
    
    <?= $form->field($model,'MontoMax') ?>

</div>
```

# Expected Output
The code should not change after beautification.

# Actual Output
```
<div class="modal-body">

    <?= $form->field($model, 'Categoria') ?>

        <?= $form->field($model, 'MontoMin') ?>

            <?= $form->field($model, 'MontoMax') ?>

</div>
```

sdvallejo:
Fixes #1339

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -318,8 +318,8 @@ function Beautifier(html_source, options, js_beautify, css_beautify) {
 
                 // Doctype and xml elements
                 '!doctype', '?xml',
-                // ?php tag
-                '?php',
+                // ?php and ?= tags
+                '?php', '?=',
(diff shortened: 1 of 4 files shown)


bitwiseman:
@sdvallejo  Thanks!

PR #1347, opened 2018-02-18
Fix multiline tags with line break before whitespace

### LINKED ISSUE #1346: Beautification of Multiline PHP
# Description
The formatting in PHP files including HTML is incorrect when there is multiline PHP code. The tag after the PHP code is indented.

# Input
```
<?php
echo "A";
echo "B";
?>
<span>Test</span>
```

# Expected Output
The code should not change after beautification.

# Actual Output
```
<?php
echo "A";
echo "B";
?>
    <span>Test</span>
```

sdvallejo:
Fixes #1346


bitwiseman:
@sdvallejo Thanks!

PR #1354, opened 2018-03-02
Improve HTML attribute wrap

mohd-akram:
This ensures that `wrap-line-length` is respected where possible.

### DIFF
--- a/test/data/html/tests.js
+++ b/test/data/html/tests.js
@@ -462,7 +462,7 @@ exports.test_data = {
         }, {
             fragment: true,
             input: '<div lookatthissuperduperlongattributenamewhoahcrazy0="true" attr0 attr1="123" data-attr2="hello    t here" heymanimreallylongtoowhocomesupwiththesenames="false">This is some text</div>',
-            output: '<div{{indent_attr_first}}lookatthissuperduperlongattributenamewhoahcrazy0="true"{{indent_attr}}attr0{{indent_attr}}attr1="123"{{indent_attr}}data-attr2="hello    t here"{{indent_over80}}heymanimreallylongtoowhocomesupwiththesenames="false"{{indent_end}}>This is some text</div>'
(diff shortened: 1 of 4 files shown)


andreyvolokitin:
Possibly related bug was addressed [here](https://github.com/beautify-web/js-beautify/pull/1329/files#diff-73474d346ba5e7ef772942fca4411708R809): ```line_char_count``` did not reset properly which caused ```wrap_line_length``` to not be respected in some cases

test/data/html/tests.js
```diff
@@ -473,8 +473,8 @@ exports.test_data = {
             output: '<?xml version="1.0" encoding="UTF-8" ?>\n<root{{indent_attr_first}}attr1="foo"{{indent_attr}}{{indent_attr_faligned}}attr2="bar"{{indent_end_selfclosing}}/>'
         }, {
             fragment: true,
-            input: '<link href="//fonts.googleapis.com/css?family=Open+Sans:300italic,400italic,600italic,700italic,400,600,700,300&amp;subset=latin" rel="stylesheet" type="text/css">',
```
bitwiseman: Rather than replacing the existing test with a different one, keep the existing test(s) and then add the new test(s).

test/data/html/tests.js
```diff
@@ -684,8 +684,8 @@ exports.test_data = {
                 output: '<div {{#if test}} class="foo" {{/if}}>^^^&content$$$</div>'
             }, {
                 fragment: true,
-                input_: '<div{{#if thing}}{{somestyle}}class="{{class}}"{{else}}class="{{class2}}"{{/if}}>^^^&content$$$</div>',
-                output: '<div {{#if thing}} {{somestyle}} class="{{class}}" {{else}} class="{{class2}}" {{/if}}>^^^&content$$$</div>'
+                input_: '<div{{#if thing}}{{somestyle}}class="{{c1}}"{{else}}class="{{c2}}"{{/if}}>^^^&content$$$</div>',
```
bitwiseman: Rather than replacing the existing test with a different one, keep the existing test(s) and then add the new test(s).

bitwiseman:
Thanks for submitting!

Issue #1156, opened 2017-03-11
Less function parameters are wrapped unexpected


shuoshubao:
**English**
As shown in the following figure, when the value of url () inside the quotation marks, the following or behind the scope of the Less function parameters will be wrapped.

**中文**
如下图所示，当url()里面的值有引号时，其下面或后面的作用域里的Less函数的参数将会换行。

**Source Code**
```
a {
  background-image: url(a.png);
}
b {
  .size(10, 20);
}


a {
  background-image: url('a.png');
  span {
    .size(10, 20);
  }
}
b {
  .size(10, 20);
}
```

**formated**

```
a {
  background-image: url(a.png);
}
b {
  .size(10, 20);
}


a {
  background-image: url('a.png');
  span {
    .size(10,
    20);
  }
}
b {
  .size(10,
  20);
}
```

![image](https://cloud.githubusercontent.com/assets/8306792/23824450/192113f2-06b2-11e7-8df6-77232297a83b.png)

PR #1359, opened 2018-03-12
issue-1156:fixed line break in mixins after background url

UGAROY:
Change-Id: I91fa77493eafb9de820eba802529dcf84ddd0170

See issue in:  https://github.com/beautify-web/js-beautify/issues/1156.

If "(" is detected, and it does not run in the eatString(')') block, parenLevel should always add 1.

### DIFF
--- a/js/lib/beautify-css.js
+++ b/js/lib/beautify-css.js
@@ -540,6 +540,7 @@ function Beautifier(source_text, options) {
                             print_string(eatString(')'));
                         } else {
                             pos--;
+                            parenLevel++;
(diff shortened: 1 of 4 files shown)


bitwiseman:
@UGAROY - 
Cool, thanks.  Look like it should work. 
Can you add a few tests for this as well?

UGAROY:
Hi, @bitwiseman 
I just added tests for the update. Plz let me know if that is ok.

bitwiseman:
Thanks!

PR #1390, opened 2018-05-08
Improve HTML attribute wrap

bitwiseman:
Reinstating #1354


bitwiseman:
This is going to be overridden but the html tokenizer.   On the plus side attribute wrapping will be come much easier.

Issue #1033, opened 2016-09-24
not correctly joining lines for HTML


garretwilson:
This is really hurting me. Here is an example HTML file:

``` html
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html>
<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">

<head>
    <title>Test Line Join</title>
</head>

<body>
    <p>The key words <q><span class="spec-must">must</span></q>, <q><span

          class="spec-must-not">must not</span></q>, <q><span class="spec-must">required</span></q>, <q><span

          class="spec-must">shall</span></q>, <q><span class="spec-must-not">shall not</span></q>, <q><span

          class="spec-should">should</span></q>, <q><span class="spec-should-not">should not</span></q>, <q><span

          class="spec-should">recommended</span></q>, <q><span class="spec-may">may</span></q>, and <q><span

          class="spec-may">optional</span></q> in this document are to be interpreted as described in <a href="#ref-rfc2119" class="ref">RFC 2119</a>.</p>
</body>

</html>
```

js-beautify should be getting rid of all that whitespace inside the `<span …>` tag. I want js-beautify to give me this:

``` html
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html>
<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">

<head>
    <title>Test Line Join</title>
</head>

<body>
    <p>The key words <q><span class="spec-must">must</span></q>, <q><span class="spec-must-not">must not</span></q>, <q><span class="spec-must">required</span></q>, <q><span class="spec-must">shall</span></q>, <q><span class="spec-must-not">shall not</span></q>, <q><span class="spec-should">should</span></q>, <q><span class="spec-should-not">should not</span></q>, <q><span class="spec-should">recommended</span></q>, <q><span class="spec-may">may</span></q>, and <q><span class="spec-may">optional</span></q> in this document are to be interpreted as described in <a href="#ref-rfc2119" class="ref">RFC 2119</a>.</p>
</body>

</html>
```

But it doesn't. This is a huge problem and is really dragging down the workflow. (The reason the lines are like that to begin with is a bug in BlueGriffon. So bugs in js-beautify are preventing it from fixing the bugs in BlueGriffon. Such is life.)

See #841.

Perhaps you could point me to a place in the code I might go to try to put in some workaround?

HookyQR:
[...]

PR #1407, opened 2018-06-01
Format inline HTML tags

madman-bob:
Fixes https://github.com/beautify-web/js-beautify/issues/1033

Replaces concept of `unformatted` tags with `inline` tags, which are not wrapped. Previous `unformatted` functionality moved to `content_unformatted` tags. This allows us to format the attributes of inline tags correctly.

### DIFF
--- a/js/lib/beautify-html.js
+++ b/js/lib/beautify-html.js
@@ -47,8 +47,9 @@
     wrap_line_length (default 250)            -  maximum amount of characters per line (0 = disable)
     brace_style (default "collapse") - "collapse" | "expand" | "end-expand" | "none"
             put braces on the same line as control statements (default), or put braces on own line (Allman / ANSI style), or just put end braces on own line, or attempt to keep them where they are.
+    inline (defaults to inline tags) - list of tags to be considered inline tags
     unformatted (defaults to inline tags) - list of tags, that shouldn't be reformatted
-    content_unformatted (defaults to pre tag) - list of tags, whose content shouldn't be reformatted
+    content_unformatted (defaults to ["pre", "textarea"] tags) - list of tags, whose content shouldn't be reformatted
     indent_scripts (default normal)  - "keep"|"separate"|"normal"
     preserve_newlines (default true) - whether existing line breaks before elements should be preserved
                                         Only works before elements, not inside tags or for text.
@@ -206,6 +207,7 @@ function Beautifier(html_source, options, js_beautify, css_beautify) {
         indent_character,
         wrap_line_length,
         brace_style,
+        inline_tags,
         unformatted,
         content_unformatted,
         preserve_newlines,
@@ -237,7 +239,7 @@ function Beautifier(html_source, options, js_beautify, css_beautify) {
     indent_character = (options.indent_char === undefined) ? ' ' : options.indent_char;
     brace_style = (options.brace_style === undefined) ? 'collapse' : options.brace_style;
     wrap_line_length = parseInt(options.wrap_line_length, 10) === 0 ? 32786 : parseInt(options.wrap_line_length || 250, 10);
-    unformatted = options.unformatted || [
+    inline_tags = options.inline || [
         // https://www.w3.org/TR/html5/dom.html#phrasing-content
         'a', 'abbr', 'area', 'audio', 'b', 'bdi', 'bdo', 'br', 'button', 'canvas', 'cite',
         'code', 'data', 'datalist', 'del', 'dfn', 'em', 'embed', 'i', 'iframe', 'img',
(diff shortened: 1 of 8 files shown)


bitwiseman:
@madman-bob 
This is very exciting. :)

One problem I see so far is this is going to be a breaking change as you've written it.  That would mean having to update the beautifier to v2 and I only want to do that as part of larger changes. 

Would you be willing to modify it to make it so it is non-breaking?  Specifically, if someone uses `unformatted` it should continue to work.

madman-bob:
Understandable. Would it be acceptable to alias `unformatted` to `unformatted_content`? It would still change the behaviour of `unformatted`, but only by formatting the attributes, when it didn't before. I can't imagine cases where you wouldn't want to format the attributes.

On an unrelated note, do you have a preferred letter for the command line name of `inline`? All the obvious ones seem to be taken.

test/data/html/node.mustache
```diff
@@ -275,18 +275,20 @@ function run_html_tests(test_obj, Urlencoded, js_beautify, html_beautify, css_be
...
                       '<div>should <custom-element>not</custom-element>' +
                       ' insert newlines</div>');
-        opts.unformatted = unformatted;
+        opts.inline_tags = inline_tags;
 
-        // Tests that don't pass, but probably should.
-        // bth('<div><span>content</span></div>');
+        bth('<div><span>content</span></div>');
+
+        test_fragment('<p>Should remove <q><span \n\nclass="some-class">attribute</span></q> newlines</p>',
```
bitwiseman: Instead of  just this one test, add a whole new section showing the accurate behavior of inlines.  
You also need to show them behaving correctly in line-wrap situations.

js/lib/beautify-html.js
```diff
@@ -237,7 +237,7 @@ function Beautifier(html_source, options, js_beautify, css_beautify) {
     indent_character = (options.indent_char === undefined) ? ' ' : options.indent_char;
     brace_style = (options.brace_style === undefined) ? 'collapse' : options.brace_style;
     wrap_line_length = parseInt(options.wrap_line_length, 10) === 0 ? 32786 : parseInt(options.wrap_line_length || 250, 10);
-    unformatted = options.unformatted || [
+    inline_tags = options.inline_tags || [
```
bitwiseman: The public option is called `inline`, not `inline_tags`.

bitwiseman:
@madman-bob 
If keeping the existing `unformatted` behavior is a serious blocker, I'm open to having it map to `inline` or `content-unformatted`, which ever produces the least changes (or at least the most sensible changes) from the existing output.

People use `unformatted` to work with xml documents and add templating elements, it may be hard to get the behavior to not break those scenarios.

test/data/html/tests.js
```diff
@@ -1010,23 +1010,23 @@ exports.test_data = {
             ]
         }, {
             fragment: true,
-            input: '<div><p>Beautify me</p></div><p><p>But not me</p></p>',
+            input: '<div><p>Beautify me</p></div><p><div>But not me</div></p>',
```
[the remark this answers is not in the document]
madman-bob: This test wasn't testing what it was intended to test.

Here, `p` is `content_unformatted`, so the final `</p>` tag is unmatched, in the same way that in

```
<textarea><textarea></textarea></textarea>
```

we have the opening tag `<textarea>`, the literal text `"<textarea>"`, followed by the closing tag `</textarea>` to match the first tag, and then the unmatched tag `</textarea>`.

madman-bob:
Have now added `unformatted` tags back in, and some more tests.

js/src/html/beautifier.js
```diff
@@ -229,6 +239,17 @@ function Beautifier(html_source, options, js_beautify, css_beautify) {
             }
         };
 
+        this.get_last_tag = function() {
+            var last_tag;
+            for (var i = this.output.length - 1; i >= 0; i --) {
```
bitwiseman: Walking back one character at a time is going to have really bad performance characteristics. 
At minimum, we should store the `last_tag` info when we find a tag and then replace it each time a new tag is found.

js/src/html/beautifier.js
```diff
@@ -182,6 +184,14 @@ function Beautifier(html_source, options, js_beautify, css_beautify) {
                     }
                 }
                 return false;
+            },
+            get_tag_name: function(full_tag) {
```
bitwiseman: This information already found during the process of processing the tag: https://github.com/madman-bob/js-beautify/blob/def3cbda7911a343ff603b81a03d63670db456a4/js/src/html/beautifier.js#L534

It would be better to store or pass that information back form there.

js/src/html/beautifier.js
madman-bob: This is walking back one *line* at a time, and, while I wrote it defensively, I cannot think of any situation where it would loop more than twice.

js/src/html/beautifier.js
madman-bob: The linked line only gets the tag name, while this function also distinguishes whether it's an opening/closing tag. Ideally, I'd refactor the other locations to use this function, but that would introduce minor behavioural changes.

bitwiseman:
@madman-bob 
Thanks!

PR #1414, opened 2018-06-09
Updated index.html

MTco:
Added [missing] doctype and possibly followed HTML5 spec in that I believe the first thing immediately after the head tag is supposed to be the charset; also added the old charset method for possible increased compatibility.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,4 +1,5 @@
-<html>
+<!doctype html>
+<html lang="en-US" dir="ltr" prefix="og:http://ogp.me/ns# fb:http://ogp.me/ns/fb#">
 <!--
   The MIT License (MIT)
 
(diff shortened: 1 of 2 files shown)


bitwiseman:
Better compat is good, but why does that include package-lock changes? 

Also, what is the `ogp.me` entries for?

MTco:
@bitwiseman package-lock changes can be attributed to my messing up the edit/revision process.

ogp.me pertains to "Open Graph namespace;" i.e., https://stackoverflow.com/questions/48933419/opengraph-prefix-in-header-necessary/50888773#50888773

bitwiseman:
@MTco Okay, so, you'll remove package-lock changes?

PR #1427, opened 2018-06-28
add static minified files

epexa:
**added minified version for jsDelivr:**
https://www.jsdelivr.com/package/npm/js-beautify

now exists only not minified files in repo:
https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js
https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify-html.js
https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify-css.js

and also not in jsDelivr:
https://cdn.jsdelivr.net/npm/js-beautify@1.7.5/js/lib/beautify.js
https://cdn.jsdelivr.net/npm/js-beautify@1.7.5/js/lib/beautify-html.js
https://cdn.jsdelivr.net/npm/js-beautify@1.7.5/js/lib/beautify-css.js


epexa:
@bitwiseman "It is not sufficient to checkin the minified current version of the file. Minification needs to be added to the build."
these minified files are copied from CloudFlare:

- https://cdnjs.cloudflare.com/ajax/libs/js-beautify/1.7.5/beautify.min.js
- https://cdnjs.cloudflare.com/ajax/libs/js-beautify/1.7.5/beautify-css.min.js
- https://cdnjs.cloudflare.com/ajax/libs/js-beautify/1.7.5/beautify-html.min.js

bitwiseman:
@epexa That is fine, but what happens when the unminified files are updated?  Will you be maintaining the minified files?  

Also, I'm not sure I understand what you're trying to do. 
https://www.jsdelivr.com/package/npm/js-beautify already has minified files: https://cdn.jsdelivr.net/npm/js-beautify@1.7.5/js/lib/beautify.min.js 

What does this change add?

bitwiseman:
We're no longer checking in the built files (minified or unminified).  We should probably add the minified files to package, but it is not required.

Issue #1254, opened 2017-09-18
Add install tests for packages


bitwiseman:
1.7.0 Release showed why full install tests are needed.  
Implement tests that the built packages will successfully install and run.

szeck87:
I would suggest adding MacOS as an OS in the Travis build as well.

bitwiseman:
@szeck87 - Didn't know they'd added that option. Perhaps you could do a pull request for that?
(See #1256)

bitwiseman:
@szeck87  Travisci's MacOS support is pretty spotty.  We now have tests that cover python and node packages.  The website is what I use when testing locally so it gets plenty of coverage.

PR #1455, opened 2018-07-21
Use entry_points to create script. Fix #1266 (script not working on Windows)

### LINKED ISSUE #1266: Generate js-beautify executable properly on windows when installed from PIP
Hi,

After installing js-beautify on a python virtualenv (i.e `pip install jsbeautifier`) one file js-beautify with no extension has been created on the scripts folder, which contains:

    #!d:\software\sublim~2\data\packages\miscel~1\test\scripts\python.exe
    #
    # Stub script to run jsbeautifier
    #
    import sys
    from jsbeautifier import main
    sys.exit(main())

This file, even if it's found in the path (`where js-beautify`) won't work on windows, it doesn't know anything neither about shebangs nor files without extension... so, could you please fix the setup.py properly so a windows executable will be created instead (like flake8, pep8, youtube-dl, ...). For more info about it take a look to here [http://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation](http://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation). One real example here [https://github.com/altendky/st/blob/lib/setup.py#L14](https://github.com/altendky/st/blob/lib/setup.py#L14).

Thanks in advance.

vivainio:


bitwiseman:
This is cool and I'd love to merge it but my python foo isn't strong enough.  Help?

vivainio:
Turned out the CI failed because there was a test that tried to use the "bad" script that was now removed. I would probably need to remove those tests as well.

The script is only created now when an installation happens, e.g. with pip install

bitwiseman:
@vivainio Is there a way to run the pip install such that the script in question gets created?

vivainio:
Yea, 'pip install .' installs it from file system

bitwiseman:
That is where it installs from.  We need to install it to a location where
we can check/test it.

bitwiseman:
Fixes #1254

PR #1488, opened 2018-08-14
Force newlines between nested rules in SCSS

### LINKED ISSUE #657: newline_between_rules support for Sass (enhancement)
Now `newline_between_rules` option in **v1.5.5** is supported for CSS only https://github.com/beautify-web/js-beautify/pull/574
So, in **Sass** for nesting doesn't work.

Here my `test.js` file:

``` js
var fs = require('fs');
var beautify_css = require('js-beautify').css;

fs.readFile('test.scss', 'utf8', function(err, data) {
  if (err) {
    throw err;
  }

  console.log(beautify_css(data, {
    indent_size: 2,
    newline_between_rules: true
  }));
});
```

Output:

``` sh
$ node test.js

.icons {
  padding: 0;
  li {
    display: inline-block;
  }
  a {
    display: block;
    color: #000;
  }
  a:hover {
    color: #ccc;
  }
}
```

I hope to add `newline_between_rules` support for **Sass**.
Regards.

### LINKED ISSUE #1258: Add or preserve empty line between nested SCSS rules
# Description
Even though I'm using `newline_between_rules`, it works only for top level (ie, zero-indented) rules (as confirmed by [this changeset](https://github.com/beautify-web/js-beautify/pull/574/files#diff-e6bf79593be5e3c81e776945fac0fdda) for the `newline_between_rules` PR). It's highly useful for SCSS code to keep newlines also for nested rules, so I think this would be a feature request. Basically, I think that if the `newline_between_rules` option is `true`, it should keep the newlines also for the nested rules. Until this gets changed, I sadly need to disable it for my SCSS code.

# Input
```scss
.list-group {
  .list-group-item {
  }
  
  .list-group-icon {
  }
}

.list-group-condensed {
}
```

# Expected Output
```scss
.list-group {
  .list-group-item {
  }

  .list-group-icon {
  }
}

.list-group-condensed {
}
```

# Actual Output
```scss
.list-group {
  .list-group-item {
  }
  .list-group-icon {
  }
}

.list-group-condensed {
}
```

# Steps to Reproduce
Beautify SCSS file with example input.

## Environment
VSCode [Beautify](https://github.com/HookyQR/VSCodeBeautify) plugin version `1.1.1`.

## Settings
Example:
```json
{
    "end_with_newline": true,
    "brace_style": "collapse,preserve-inline",
    "max_preserve_newlines": 2,
    "newline_between_rules": true,
    "space_around_combinator": true,
    "space_after_anon_function": true,
    "wrap_line_length": 100,
    "css": {
        "indent_size": 2
    }
}
```

MacKLess:
Fixes #1258 
Fixes #657

### DIFF
--- a/js/src/core/output.js
+++ b/js/src/core/output.js
@@ -269,5 +269,19 @@ Output.prototype.just_added_blankline = function() {
   return false;
 };
 
+Output.prototype.ensure_empty_line_above = function(commentPattern) {
+  var index = this._lines.length - 2;
+  while (index >= 1) {
+    var potentialEmptyLine = this._lines[index];
+    if (potentialEmptyLine.is_empty()) {
+      break;
+    } else if (potentialEmptyLine._items[0].indexOf(commentPattern) !== 0) {
+      this._lines.splice(index + 1, 0, new OutputLine(this));
+      this.previous_line = this._lines[this._lines.length - 2];
+      break;
+    }
+    index--;
+  }
+};
 
 module.exports.Output = Output;
--- a/js/src/css/beautifier.js
+++ b/js/src/css/beautifier.js
@@ -285,8 +285,10 @@ function Beautifier(source_text, options) {
           eatWhitespace(true);
           output.add_new_line();
 
-          if (newline_between_rules && indentLevel === 0 && !output.just_added_blankline()) {
-            output.add_new_line(true);
+          if (newline_between_rules && !output.just_added_blankline()) {
+            if (input.peek() !== '}') {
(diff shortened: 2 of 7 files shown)


bitwiseman:
@MacKLess 
There's some test inputs that aren't what people expected.  That said, this is a huge step forward.  

Getting this to the point of have a new line **before** all  rules instead of **between** rules may be more work than is reasonable to expect. 

I can wait to merge this, or take it as it is as a significant improvement.  What do you think?

Issue #1495, opened 2018-08-22
Publish v1.8.0


bitwiseman:
@garretwilson @amandabot @HookyQR @astronomersiva 
@madman-bob @swan46 @LoganDark @MacKLess 
@Mlocik97-issues @Hirse @jdavisclark

At this point the 1.8.0 release includes at least 80 fixed bugs and enhancements.  HTML and CSS/LESS/SCSS beautification have both seen massive improvements.  Not that performance is primary focus of this project, but the HTML beautifier in particular has seen 10x or more performance improvement. 

I'd like to tie off this RC cycle and publish 1.8.0 (release). The latest 1.8.0-rc12 is up on http://jsbeautifier.org/  and npm and pypi.

If you can try out some of the inputs you use most or otherwise give this RC a sanity test drive, I'd be most grateful.  

I plan to release soon.

LoganDark:
Seems good.

garretwilson:
I'm very excited about this. I hadn't been wanting to rush you, but I've been watching the release candidates with anticipation. You know that I'm anxious to get the fix for #1033 published.

I updated atom-beautify to the latest v0.33.0, and then I tried to follow my own directions at #1033 for testing that fix. According to my notes, I copied the entire contents of `js/lib` in the js-beautify distribution to `~/.atom/packages/atom-beautify/node_modules/js-beautify/js/lib`, replacing the atom-beautify files there.

But now there doesn't seem to be a `js/lib` directory anymore, neither in the latest `js-beautify-1.8.0-rc12.zip` nor in the commit pull from `master` (3374af3).

bitwiseman:
@garretwilson 

In the final release the atom plugin will get the  files from the npm package: https://registry.npmjs.org/js-beautify/-/js-beautify-1.8.0-rc12.tgz

There no change for any normal usage. We only stopped checking in the built files in the master branch as they are not interesting and they were was causing confusion for contributors trying to figure out what files they should edit.

Hirse:
For the Brackets extension I usually grab the files from that folder because I don't need the rest of the package.
Would it be possible to also add them to the [release](https://github.com/beautify-web/js-beautify/releases) on GitHub?

astronomersiva:
Thanks a lot for your continuous efforts on this @bitwiseman!

I ran into a issue while trying out 1.8.0-rc12. I have submitted an [issue](https://github.com/beautify-web/js-beautify/issues/1496) for this.

[...]

PR #1499, opened 2018-08-22
Fixed whitespace detection issue with tests

### LINKED ISSUE #1496: ReferenceError: token is not defined
# Description
I am getting the error message `ReferenceError: token is not defined` while running this on text that has [NO-BREAK-SPACE character](https://www.fileformat.info/info/unicode/char/00a0/index.htm) in the content.

# Input
```
<p>hello world</p>
```

# Expected Output
```
<p>hello world</p>
```

# Actual Output
```
ReferenceError: token is not defined
    at Beautifier.beautify (/Users/siva/.nvm/versions/node/v8.11.3/lib/node_modules/js-beautify/js/src/html/beautifier.js:372:29)
    at style_html (/Users/siva/.nvm/versions/node/v8.11.3/lib/node_modules/js-beautify/js/src/html/index.js:33:21)
    at Function.style_html (/Users/siva/.nvm/versions/node/v8.11.3/lib/node_modules/js-beautify/js/src/index.js:36:10)
    at makePretty (/Users/siva/.nvm/versions/node/v8.11.3/lib/node_modules/js-beautify/js/lib/cli.js:448:36)
    at Object.processInputSync (/Users/siva/.nvm/versions/node/v8.11.3/lib/node_modules/js-beautify/js/lib/cli.js:442:5)
    at Array.forEach (<anonymous>)
    at Object.exports.interpret (/Users/siva/.nvm/versions/node/v8.11.3/lib/node_modules/js-beautify/js/lib/cli.js:305:15)
    at Object.<anonymous> (/Users/siva/.nvm/versions/node/v8.11.3/lib/node_modules/js-beautify/js/bin/js-beautify.js:4:5)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
```

js-beautify version - 1.8.0-rc12

MacKLess:
Fixes #1496
@astronomersiva this converts the &nbsp character into a space. I think this is a decent fix, because people generally should be using `&nbsp;` in html. Let me know if you disagree.

### DIFF
--- a/js/src/core/tokenizer.js
+++ b/js/src/core/tokenizer.js
@@ -43,8 +43,8 @@ var Tokenizer = function(input_string, options) {
   this.__newline_count = 0;
   this.__whitespace_before_token = '';
 
-  this._whitespace_pattern = /[\n\r\u2028\u2029\t ]+/g;
-  this._newline_pattern = /([\t ]*)(\r\n|[\n\r\u2028\u2029])?/g;
+  this._whitespace_pattern = /[\n\r\u2028\u2029\t\u000B\u00A0\u1680\u180e\u2000-\u200a\u202f\u205f\u3000\ufeff ]+/g;
+  this._newline_pattern = /([^\n\r\u2028\u2029]*)(\r\n|[\n\r\u2028\u2029])?/g;
 };
 
 Tokenizer.prototype.tokenize = function() {
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -369,7 +369,7 @@ Beautifier.prototype.beautify = function() {
       parser_token = this._handle_text(printer, raw_token, last_tag_token);
     } else {
       // This should never happen, but if it does. Print the raw token
-      printer.add_raw_token(token);
+      printer.add_raw_token(raw_token);
     }
 
     last_token = parser_token;
--- a/js/src/html/tokenizer.js
+++ b/js/src/html/tokenizer.js
@@ -53,9 +53,12 @@ var Tokenizer = function(input_string, options) {
   BaseTokenizer.call(this, input_string, options);
   this._current_tag_name = '';
 
+  this._whitespace_pattern = /[\n\r\t ]+/g;
+  this._newline_pattern = /([^\n\r]*)(\r\n|[\n\r])?/g;
+
   // Words end at whitespace or when a tag starts
   // if we are indenting handlebars, they are considered tags
-  this._word_pattern = this._options.indent_handlebars ? /[\s<]|{{/g : /[\s<]/g;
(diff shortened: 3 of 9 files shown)


Hirse:
In HTML, I could see an argument for keeping a &nbsp character or changing it to `&nbsp;` as it can technically change the layout, even though it is certainly bad style.

garretwilson:
> In HTML, I could see an argument for keeping a &nbsp character or changing it to `&nbsp;`…

I don't know the details of this pull request; I'm only commenting because I'm participating in #1495.

* There is nothing wrong with the `U+00A0`. It is necessary in many instances, e.g. to prevent icon+menu items from wrapping. _Please never, ever change `U+00A0` to a normal space character!_ This is corrupting content, and is the worse thing you could do.
* Please do not change `U+00A0` to `&nbsp;`. While this named character reference is guaranteed to be recognized as HTML, if the source file is XML-based (e.g. XHTML) then it will break parsing because in XML all entities other than the predefined entities must be defined in an internal or external XML DTD.

So please no workarounds that involve mucking with the content. **_Leave the content as it is!!!_** A non-breaking space should be formatting according to the semantics: i.e. don't wrap on a non-breaking space.

bitwiseman:
@garretwilson 
Okay, so being strict and accurate with the whitespace detection in HTML matters.  Good to know.  Great user-feedback as always.   

@MacKLess  It looks like the HTML tokenizer will need to specifically only match "space characters" 
as noted in https://www.w3.org/TR/html5/infrastructure.html#space-characters when separating words. All other characters will need to be treated as text.

garretwilson:
> It looks like the HTML tokenizer will need to specifically only match "space characters" …

I saw the preview of your comment come up in my email, so I was going to paste the link to https://www.w3.org/TR/html52/infrastructure.html#space-characters and then I see you already found it. 😆

So 👍 !

garretwilson:
> Okay, so being strict and accurate with the whitespace detection in HTML matters.

"… being strict and accurate … matters." Thought of the day. 😉

garretwilson:
There is I suppose the question of whether you should wrap on characters that Unicode has marked as "whitespace" in the Unicode Character Database. For example the `MEDIUM MATHEMATICAL SPACE` character `U+205F` is marked as Unicode "whitespace". The problem with wrapping on Unicode whitespace characters is the same one as NBSP: when you wrap, I presume js-beautify would _replace_ the character with a newline (and perhaps tabs). The problem is that doing so would result in removal of this character with special semantics. The other option would be to keep it and still wrap, which would result in the _addition_ of ASCII whitespace that wasn't there before.

So after this short little conversation with myself 😆 I think my earlier conclusion (and that of @bitwiseman ) is the correct one: only wrap on the characters HTML5 considers whitespace, and we'll all be happy.

Thank you for bearing with this brief interlude. 😉

PR #1525, opened 2018-09-01
Add indent level options functionality with tests

### LINKED ISSUE #724: beautify.js doesn't have indent_level option
In the [options documentation](https://github.com/beautify-web/js-beautify#options) I found the setting `indent_level` as _Initial indentation level_ for both CLI and underscored option.

However when using it in the js version, the setting seems to be ignored.
When checking the code, it seems like it should be included somewhere [here](https://github.com/beautify-web/js-beautify/blob/master/js/lib/beautify.js#L282), but it isn't.

MacKLess:
Fixes #724

This fix addresses most of the functionality with one major exception. If you set the indent level to zero in CSS and JS, the beautifier still autodetects the indentation of the inputted code. I will be filing a bug for this specific functionality.

### DIFF
--- a/js/src/css/beautifier.js
+++ b/js/src/css/beautifier.js
@@ -174,13 +174,11 @@ Beautifier.prototype.beautify = function() {
 
   // reset
   var baseIndentString = '';
-  var preindent_index = 0;
-  if (source_text && source_text.length) {
-    while ((source_text.charAt(preindent_index) === ' ' || source_text.charAt(preindent_index) === '\t')) {
-      preindent_index += 1;
-    }
-    baseIndentString = source_text.substring(0, preindent_index);
-    source_text = source_text.substring(preindent_index);
+  if (this._options.base_indent_string) {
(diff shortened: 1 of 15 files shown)


js/src/css/beautifier.js
```diff
@@ -174,13 +174,11 @@ Beautifier.prototype.beautify = function() {
...
-  if (source_text && source_text.length) {
-    while ((source_text.charAt(preindent_index) === ' ' || source_text.charAt(preindent_index) === '\t')) {
-      preindent_index += 1;
-    }
-    baseIndentString = source_text.substring(0, preindent_index);
-    source_text = source_text.substring(preindent_index);
+  if (this._options.base_indent_string) {
+    baseIndentString = this._options.base_indent_string;
+  } else {
+    var match = source_text.match(/^[\t ]*/);
```
bitwiseman: It would be better if these were refactored into a common function but the html beautifier would not be able to use it. 😢 So it goes.

PR #1559, opened 2018-09-27
Fix extendscript #include

### LINKED ISSUE #1114: js-beautify adds a space between # and include
I'm using js-beautify when coding for Adobe After Effects that uses Javascript. The problem is that when I write #include js-beautify changes it to # include and breaks the code.
Is there a way to ignore # signs or another way to fix this or will it need to be done in the source-code?

Thanks!

MacKLess:
Fixes #1114


bitwiseman:
@MacKLess 
I made minor changes to make your fix work with different types of newlines.  
Could you take a look at the additional refactor I did to shebang processing and tell me if you agree?  (It passes tests, but I'd like your thumbs up.)

MacKLess:
@bitwiseman Thanks for beefing up the testing section. I completely forgot about u2028 and u2029. Looks great to me!

Issue #1446, opened 2018-07-09
Create unit tests for Core 


bitwiseman:
There are no unit tests for core classes.

ganes1410:
hey @bitwiseman . I will try to do that.Can you help me get started ?

bitwiseman:
@ganes1410  
Here is the first unittest for InputScanner.  
https://github.com/beautify-web/js-beautify/commit/54f8e464b4b375640710955d9f1898f6cd5220e2 

You can add more for this class and the others in Core.  Thanks!

ganes1410:
@bitwiseman ok.i'll start working on this.

bitwiseman:
Run `make jstest` and/or `make pytest`.

ganes1410:
ok.Thanks @bitwiseman.can i use mocha.It's already present in the package.json

bitwiseman:
Yes, that is the intent.

ganes1410:
@bitwiseman  Sorry.I asked why can't we run tests by directly running `mocha`

bitwiseman:
@ganes1410 You can do that too.

ganes1410:
@bitwiseman  I didn't find time to work on it . I will try to complete it by this weekendb

bitwiseman:
You don't have do _all_ of them.  Any number of them would be great.  😄

athiramanu:
Hi @bitwiseman
I would like to contribute some tests to this task

codewithmitesh:
Hey, @bitwiseman  I'd want to add a few more tests to this. Is it available?

bitwiseman:
Go for it.

xavya-dev:
@bitwiseman - this issue is still open can i please work on this and add more test?

Issue #1364, opened 2018-03-31
Add function default_options() to beautifier.js


tkrotoff:
You can retrieve Beautifier default options using Python: `opts = jsbeautifier.default_options()`, not with JavaScript.

This would be useful for "complex" default options, example with [beautifier.js source code for `unformatted`](https://github.com/beautify-web/js-beautify/blob/v1.7.5/js/src/html/beautifier.js#L93):

```JavaScript
// beautifier.js simplified source code for unformatted
// https://github.com/beautify-web/js-beautify/blob/v1.7.5/js/src/html/beautifier.js#L93
defaultOptions.unformatted = [
  'a', 'abbr', 'area', 'audio', 'b', 'bdi', 'bdo', 'br', 'button', 'canvas', 'cite'
  // ...
];
```

If I want to remove (or add) some tags from the unformatted option I would like to write something like:
```JavaScript
import { html_beautify, defaultOptions } from 'js-beautify';

const excludeButton = defaultOptions.unformatted.filter(tag => tag !== 'button');
const niceHtml = html_beautify(uglyHtml, { unformatted: excludeButton });
```

bitwiseman:
Fixed by #1568.

PR #1568, opened 2018-10-06
defaultOptions for js

zilioti:
Hi @bitwiseman,
since I worked in the tests for options core in #1446, I decided to take a chance at doing #1364.

I opened this PR with the basic idea to check if the below implementation is desired by @tkrotoff: 
```javascript
import { html_beautify } from 'js-beautify';

const excludeButton = html_beautify.defaultOptions.unformatted.filter(tag => tag !== 'button');
const niceHtml = html_beautify(uglyHtml, { unformatted: excludeButton });
```

Also, could you guide me where should I test? I made just a basic one for now.


bitwiseman:
@zilioti 
The `defaultOptions` should probably be functions that return new instances of each Options when called.  Otherwise someone could change the defaultOptions and side-effect unexpectedly.

Other than that this seems great.

bitwiseman:
@zilioti 
Also this will need be exposed via the legacy packages.

PR #1572, opened 2018-10-12
Tab and space alignment correction

### LINKED ISSUE #1294: Using tabs when wrapping attributes and wrapping if needed
I was wondering if you're open to receiving two PRs to add two options for HTML.

1) An option to use tabs when wrapping parameters + indent_with_tabs:true. Right now it adds spaces, creating a mixed mode file. Something like `wrap_attributes_with_tabs:bool`.

2) An option to wrap parameters if they reach the wrap_line_length. Something similar to force-expand-multi-line but not forced. I don't have a good name for this option because I don't understand quite well what `auto` means for `wrap_attributes`. 

We'd end up with a file formatted this way:

```html
<div
	id="main"
	class="outer"
	data-one="long text to wrap"
	data-two="long text to wrap"
>
	<div id="sub" class="inner">
		content
	</div>
</div>
```

This is to match a common JS/TS style used when wrapping parameters:

```ts
public test(
	oneVeryLongParameter: string,
	twoLongParameters: string,
	threeLongParameters: string
) {
	console.log("me");
}
```

Thanks

MacKLess:
Fixes #1294

This forces alignment to use tabs when indicated or defaults to spaces. 

If you look at this line in the js.test file (https://github.com/beautify-web/js-beautify/pull/1572/files#diff-e21e76a3df495fbbe325a76d7e3431eaR368), you'll see that what was originally 7 spaces, is now processed as a tab + 3 spaces.

This is a breaking change that will affect people who use any of the "aligned" options; however, this is correct functionality.


bitwiseman:
This is I highly requested feature.  
Now that we've released 1.8.7,  I'm willing to start a v1.8.8-alpha series  to let people give it a try.

PR #1577, opened 2018-10-22
Document the default value of "editorconfig"

markstos:
Also show that it's a valid option in the  config file.


bitwiseman:
A missing setting defaults to `falsey`.  It isn't interesting to add documentation for all the false settings.

markstos:
First, I've used a lot configuration systems over the years and I can't recall where where it was assumed that missing values defaulted to false.

Second, it's *critical* to have a comprehensive list of configuration options so people know that they actually exist.

I was not certain that `editorconfig` was configurable until I read the source code to confirm.

PR #1580, opened 2018-10-27
Added support for globs.

### LINKED ISSUE #787: Support globs in CLI
Support e.g. `html-beautify -r *.html` (or `htm` in my case) to replace all matching files in a directory.
Also, the commands `js-beautify -r`, `html-beautify -r` and `css-beautify -r` could default to replacing `*.js`, `*.htm?(l)` and `*.css` respectively.
(Use [glob](https://www.npmjs.com/package/glob))

StefanKoenen:
Fixes #787


StefanKoenen:
@bitwiseman 
I see you reviewed another pull request, so maybe you can help me out whats going on at travis? I want to fix it so my PR can be merged, but I'm not sure what the error is?

README.md
```diff
@@ -317,7 +317,8 @@ CSS Beautifier Options:
   -n, --end-with-newline             End output with newline
   -L, --selector-separator-newline   Add a newline between multiple selectors
   -N, --newline-between-rules        Add a newline between CSS rules
-
+  -g  --glob                         Match files using a glob
```
bitwiseman: I don't think this needs an option.  It should just be enabled for all cases.

js/src/cli.js
```diff
@@ -33,23 +33,25 @@
 */
 /*jshint strict:false */
 
-var debug = process.env.DEBUG_JSBEAUTIFY || process.env.JSBEAUTIFY_DEBUG ? function() {
```
bitwiseman: This is probably why you're getting failures.  Run `make` locally and then be sure to commit any changed files.
The build runs the beautifier on the javascript files in project, enforcing code style.

js/src/cli.js
```diff
@@ -33,23 +33,25 @@
...
+var debug = process.env.DEBUG_JSBEAUTIFY || process.env.JSBEAUTIFY_DEBUG ? function () {
```
bitwiseman: This is probably why you're getting failures.  Run `make` locally and then be sure to commit any changed files.
The build runs the beautifier on the javascript files in project, enforcing code style.

js/src/cli.js
```diff
@@ -602,10 +607,31 @@ function checkFiles(parsed) {
     parsed.replace = true;
   }
 
+  // support glob
+  var hasGlob = parsed.glob !== undefined;
+
+  if (hasGlob && parsed.files.length) {
+    // Filter the files with glob (also required for a git precommit hook)
+    miniMatch = new Minimatch(parsed.glob);
```
bitwiseman: Rather than minimatching, what do you think of what I said above?  Expanding the files array with globs, check if the result is empty, and finally removing duplicates.

StefanKoenen: For all comments:
**Current situation:**
I'm using Prettier to format TypeScript code, but Prettier doesn't support HTML, so I want to use beautify-js for this, but for the HTML files only!
The --type html doesn't filter HTML files only? (or did I do something wrong?)

**Sugestion:**
After this PR is finished I want to make a pre-commit option (without the need for other packages like glob-run as sugested [here](https://github.com/beautify-web/js-beautify/issues/787#issuecomment-315628380).
The files from the pre-commit option will be added before this minimatch filter, so we are still be able to run a filter on these files. (but again, maybe I did something wrong with the --type option)

What do you think of this or how can we handle my scenario?

StefanKoenen:
@bitwiseman thanks a lot for all help, I had a look at the python script but I think it's next level for me, so some help would be very nice!

js/src/cli.js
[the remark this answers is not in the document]
bitwiseman: @StefanKoenen  Take a look at 
https://github.com/beautify-web/js-beautify/blob/master/CONTRIBUTING.md

I'm sure it needs clarification and improvement.  Any help you want to provide there is welcome.

js/src/cli.js
bitwiseman: Well, if globbing works the way I'd expected, you'd call:
```
js-beautify --type html  *.html *.htm 
```

But I get the sense you had a different idea.   Could you give some examples?

js/src/cli.js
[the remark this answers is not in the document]
StefanKoenen: Thanks this was exactly were I was looking for, not sure how I missed it 😅

README.md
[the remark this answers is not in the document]
StefanKoenen: as discussed, removed the option

js/src/cli.js
```diff
@@ -110,6 +112,7 @@ var path = require('path'),
     "version": Boolean,
     "help": Boolean,
     "files": [path, Array],
+    "glob": String,
```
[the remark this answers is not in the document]
StefanKoenen: Removed the option

js/src/cli.js
```diff
@@ -590,7 +591,17 @@ function checkFiles(parsed) {
...
-        parsed.files.push(path.resolve(f));
+        var foundFiles = glob(f, {
+          sync: true,
+          absolute: true,
+          ignore: ['**/node_modules/**', '**/.git/**']
+        });
+
+        if (foundFiles && foundFiles.length) {
+          foundFiles.forEach(function(file) {
+            parsed.files.push(file);
```
bitwiseman: @StefanKoenen 
We also need to remove or prevent duplicates. 
In the degenerate case, `src/**/*.js src/**/*.js` should not beautify files twice.

package.json
```diff
@@ -47,6 +47,8 @@
   "dependencies": {
     "config-chain": "~1.1.5",
     "editorconfig": "^0.15.0",
+    "glob": "^7.1.3",
+    "minimatch": "^3.0.4",
```
bitwiseman: Remove this.

README.md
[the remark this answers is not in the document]
bitwiseman: But.. this is still here. :)

package.json
StefanKoenen: done.

js/src/cli.js
StefanKoenen: You are right! Added a check for this.

README.md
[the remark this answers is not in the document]
StefanKoenen: Removed now :)

bitwiseman:
@StefanKoenen 
This looks much better.  The tests are failing because trying to beautify a file that doesn't exist should give a useful error message.  https://travis-ci.org/beautify-web/js-beautify/jobs/451505187#L1653

So, the glob code needs to check if the item has glob pattern chars (https://github.com/isaacs/node-glob#globhasmagicpattern-options) and then expand it.  If it isn't a pattern and it doesn't find the file, it is an error.  If there are only patterns and they return no files, I'm not sure if that should be a non-zero exit code.  Probably. 

Thanks for working on this.

StefanKoenen:
@bitwiseman 
Added check for glob pattern. If it isn't a glob pattern, it is testing the filepath. (I don't think empty glob result should return exit 0)

bitwiseman:
@StefanKoenen  
Looking good. Can you add some tests to https://github.com/beautify-web/js-beautify/blob/master/js/test/shell-test.sh ?  After that, I'll get the python side implemented and we'll get this merged.

StefanKoenen:
@bitwiseman uhm yeah I think I can :). Is it also possible to run a single test? (because it's new for me)

crhistianramirez:
How can I use this? I tried adding this npm script:
"prettyHtml": "js-beautify --html '**/*.html' -r"

Also with html-beautify:
"prettyHtml": "html-beautify '**/*.html' -r",

bitwiseman:
Try putting the `-r` before the file glog, or using the `-f` option:

```
"prettyHtml": "js-beautify --html -r '**/*.html'"
```
or 
```
"prettyHtml": "js-beautify --html -f '**/*.html' -r"
```

Issue #1616, opened 2019-01-10
DeanEdward python unpacker offset problem


mFaou:
# Description
Hi,
The unpacker python script for DeanEdward packer doesn't correctly handle the begin and end offsets.
The result is that the unpacked code is not put at the right offset in the final script. Thus, it breaks the original script.
At line 28, it searches for the beginning offset with 
`source.replace(' ', '').find('eval(function(p,a,c,k,e,')`. However at line 30 when it copies the code with `beginstr = source[:mystr]`, it doesn't take into account the number of spaces. Thus, the unpacked code is put at a "random" place.
For the end offset, the script starts the search at the beginning of the source code while it should be started after the begin offset.

# Input
The code looked like this before beautification:
```
function test (){alert ('This is a test!!!')}; eval(function(p,a,c,k,e,r){e=String;if(!''.replace(/^/,String)){while(c--)r[c]=k[c]||c;k=[function(e){return r[e]}];e=function(){return'\\w+'};c=1};while(c--)if(k[c])p=p.replace(new RegExp('\\b'+e(c)+'\\b','g'),k[c]);return p}('0(\'1\');',2,2,'alert|test'.split('|'),0,{}))
```

# Expected Output
The code should have looked like this after beautification:
```
function test() {
    alert('This is a test!!!')
};
alert('test');
```

# Actual Output
The  code actually looked like this after beautification:
```
function test() {
    alert('This is a test!alert(\'test\');
```
# Steps to Reproduce
Use the last version (1.8.9) of jsbeautifier python version (the javascript seems correct)

## Environment
OS: Linux
jsbeautifier 1.8.9

## Settings
Default

# Solution:
We can use a regex instead of the str.find function. This modified version of unpack.py fixes the bug:
```
#
# Unpacker for Dean Edward's p.a.c.k.e.r, a part of javascript beautifier
# by Einar Lielmanis <einar@beautifier.io>
#
#     written by Stefano Sanfilippo <a.little.coder@gmail.com>
#
# usage:
#
# if detect(some_string):
#     unpacked = unpack(some_string)
#

"""Unpacker for Dean Edward's p.a.c.k.e.r"""

import re
import string
from jsbeautifier.unpackers import UnpackingError

PRIORITY = 1


def detect(source):
    global beginstr
    global endstr
    beginstr = ''
    endstr = ''
    begin_offset = -1
    """Detects whether `source` is P.A.C.K.E.R. coded."""
[...]

PR #1617, opened 2019-01-11
Fix for the DeanEdward unpacker. 

mFaou:
The begin and end offsets were not correctly handled for non-minified scripts (issue #1616).

### DIFF
--- a/python/jsbeautifier/unpackers/packer.py
+++ b/python/jsbeautifier/unpackers/packer.py
@@ -24,19 +24,22 @@ def detect(source):
     global endstr
     beginstr = ''
     endstr = ''
+    begin_offset = -1
     """Detects whether `source` is P.A.C.K.E.R. coded."""
-    mystr = source.replace(' ', '').find('eval(function(p,a,c,k,e,')
-    if(mystr > 0):
-        beginstr = source[:mystr]
-    if(mystr != -1):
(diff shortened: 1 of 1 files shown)


bitwiseman:
@mFaou  
Thanks for creating this!
It looks like that failed other tests. I would love to put this fix into the next version but it will need to pass the tests.  Can I help you figure how to get this working?

mFaou:
Sorry, I didn't run the tests. I'll try to modify the script to make them pass ;)

mFaou:
Could you help me figure out why it has failed again? (for the Azure Pipelines, for codeclimate I already made a new persion pep8 compliant that I will push later).
When I run `make py` and `make pytest` on my local machine, it passes the tests.

bitwiseman:
@mFaou 
The last failures were my fault.   Should be fixed now.

bitwiseman:
@mFaou  
One more request - could you add a test that covers this issue?  
See https://github.com/beautify-web/js-beautify/blob/master/python/jsbeautifier/unpackers/tests/testpacker.py#L31-L35

You already have it in the issue you files, so it should be straightforward.  

Thanks!

mFaou:
Done. I had to modify a little bit the `check` function in `test_unpack` because the begin and end offsets are filled in the detect function. So it is not possible to use `unpack` without calling `detect` first.
If this is not the behaviour you want, we need to add a check whether beginstr and endstr are null and if so, calling detect at the beginning of unpack.

mFaou:
So, last commit to fix the pep8 compliance. Sorry for the high number of commits ;)

bitwiseman:
@mFaou  
Number of commits no problem.  Thank you for the high-quality PR.

Issue #1527, opened 2018-09-01
Change css tests to use 4 space indenting instead of tabs


bitwiseman:
# Description
The use of tabs for indenting in the css beautifier is a continual drag, making tests hard to read and prone to errors. Since indenting is now handled by the Output class for all beautifiers, it makes sense to switch to the same defaults.

PR #1650, opened 2019-03-25
Future contribution

HanabishiRecca:
Here will be some contribution from me :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -281,6 +281,12 @@
         "string-width": "^2.0.0"
       }
     },
+    "ansi-colors": {
+      "version": "3.2.3",
+      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-3.2.3.tgz",
+      "integrity": "sha512-LEHHyuhlPY3TmuUYMh2oz89lTShfvgbmzaBcxve9t/9Wuy7Dwf4yoAKcND7KFT1HAQfqZ12qtc+DUrBMeKF9nw==",
+      "dev": true
+    },
     "ansi-regex": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.0.tgz",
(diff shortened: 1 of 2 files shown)


HanabishiRecca:
https://github.com/beautify-web/js-beautify/issues/1527

python/jsbeautifier/javascript/options.py
```diff
@@ -82,6 +82,7 @@ def __init__(self, options=None):
         self.e4x = self._get_boolean('e4x')
         self.comma_first = self._get_boolean('comma_first')
         self.operator_position = self._get_selection('operator_position', OPERATOR_POSITION)
+        self.indent_empty_lines = self._get_boolean('indent_empty_lines')
```
bitwiseman: This shouldn't be needed.  It is done in `core/options.py`.

HanabishiRecca:
This is fix for https://github.com/beautify-web/js-beautify/issues/1651

HanabishiRecca:
@bitwiseman, I think you should merge this at the current state to avoid future conflicts because of lots of changes in tests.

js/src/javascript/beautifier.js
```diff
@@ -1412,12 +1413,6 @@ Beautifier.prototype.handle_dot = function(current_token) {
       this._flags.last_token.text === ')' && this._options.break_chained_methods);
   }
 
-  // Only unindent chained method dot if this dot starts a new line.
-  // Otherwise the automatic extra indentation removal will handle the over indent
-  if (this._options.unindent_chained_methods && this._output.just_added_newline()) {
-    this.deindent();
-  }
-
```
bitwiseman: This change makes me a bit nervous - it effectively makes a local setting for dots into a broader setting that could have side effects in a number of places.  

I'm going to take all the other changes in this PR now though.  Thanks!

HanabishiRecca: @bitwiseman, sorry for no explanation about this. The change is not related only to the dot. Problem is really deeper. I've traced behavior and found a bunch of problems with deindention, not with dot only. There is wrong behavior with not closed block statements within an expression.

For example:
(**unindent_chained_methods** is ON)
```
getNum(() => {
    return 10;
}) -
5;
```
Expected output:
```
getNum(() => {
    return 10;
}) -
5;
```
Actual output:
```
getNum(() => {
        return 10;
    }) -
    5;
```

So my fix works for all cases, not only for dots. I can understand your nervous because this is pretty not obvious and complicated :)

bitwiseman: @HanabishiRecca 
Your explanation  doesn't include a dot, but the change is based on `unindent_chained_methods` ... So, that means that even though the example doesn't include chained methods the behavior will change depending on whether `unindent_chained_methods` is true or false... 

You arrived because of #1651  right?

HanabishiRecca: Yes. I can agree this is a bit weird. If we want to separate this behavior, this logic really will be wrong. Ok, I will try to find another solution.

bitwiseman:
@HanabishiRecca 
Thanks again for contributing!

PR #1656, opened 2019-04-07
stop reformatting valid css \! into invalid \ !

averydev:
# Description

Fixes Issue: 

CSS allows escape of the `!` character in class names i.e. `.blue\!`, which would be applied to

```<div class="blue!" />```

However, the logic to put a space before `!important` resulted in generating invalid css `.blue\! {` turned into `.blue\ ! {`

# Before Merge Checklist 

N/A --- Added command-line option(s) (NA if -- NA
N/A --- README.md documents new feature/option(s) -- NA

### DIFF
--- a/js/src/css/beautifier.js
+++ b/js/src/css/beautifier.js
@@ -435,7 +435,7 @@ Beautifier.prototype.beautify = function() {
       if (whitespaceChar.test(this._ch)) {
         this._ch = '';
       }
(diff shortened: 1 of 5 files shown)


averydev:
I had some trouble getting the generated tests to create the output correctly, which needed a double `\\` hence the ugly `\\\\` but i didn't want to mess with the test generator extensively.

Issue #1641, opened 2019-03-18
Beautify inserts space before exclamation mark in comment <!-- in css <style>


bitwiseman:
# Description
see https://github.com/brackets-beautify/brackets-beautify/issues/278 

What I mean:
```html
<style type="text/css"><!--
--></style>
```

What happens:
```html
<style type="text/css">< !--
--></style>
```

What should happen:
```html
<style type="text/css"><!--
--></style>The input,
```

PR #1662, opened 2019-04-16
Comment wrapping

bitwiseman:
# Description

Fixes Issue: 
#1641

# Before Merge Checklist 

### DIFF
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -442,10 +442,12 @@ Beautifier.prototype._handle_text = function(printer, raw_token, last_tag_token)
 Beautifier.prototype._print_custom_beatifier_text = function(printer, raw_token, last_tag_token) {
   var local = this;
   if (raw_token.text !== '') {
-    printer.print_newline(false);
+
     var text = raw_token.text,
       _beautifier,
-      script_indent_level = 1;
+      script_indent_level = 1,
+      pre = '',
+      post = '';
     if (last_tag_token.custom_beautifier_name === 'javascript' && typeof this._js_beautify === 'function') {
       _beautifier = this._js_beautify;
     } else if (last_tag_token.custom_beautifier_name === 'css' && typeof this._css_beautify === 'function') {
@@ -456,7 +458,6 @@ Beautifier.prototype._print_custom_beatifier_text = function(printer, raw_token,
         return beautifier.beautify();
       };
     }
-
 
     if (this._options.indent_scripts === "keep") {
       script_indent_level = 0;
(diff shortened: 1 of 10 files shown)


bitwiseman:
@HanabishiRecca 
If you have a minute to look this over, that'd be great.

HanabishiRecca:
Comments formatting? I don't think this is a good idea because
1. Comments can be used for some decorative purposes. This means content must stay untouched.
Example:
```
<!--
                ._______________.
                |               |
                |   ATTENTION   |
                |_______________|

Please, keep the code clean, don't use things like this
<style>.a{}.b{}.c{}.d{}</style>

Use formatting!
<style>
    .a {}
    .b {}
    .c {}
    .d {}
</style>
-->
```
2. Using html comments inside the script and style blocks is a very bad practice.

So I think best way is completley ignore comments.

bitwiseman:
@HanabishiRecca  
Ah, I should have provided context. 
Yes, putting html comments in style and script is not recommended,  but it is supported and people do it.  😭 

I see your point about not formatting what is inside comments.  
However, when they are used inside script and style tags, they have specific and well defined behavior and the stuff inside the comment needs to be treated as style or script of the appropriate type:

https://www.w3.org/TR/REC-html40/interact/scripts.html#h-18.3.2
https://www.w3.org/TR/CSS21/syndata.html#comments

This change doesn't perfectly model those two specification, but it is better and cleaner than what is there now, fixing #1641.  It also fixes the html tokenizer to correctly handle this input (which it would fail on before): 

```
<script>
    <!--
    var a = "</script>";
    // before this change the html beautifier would think that the script tag ended on the line above
    -->
</script>
```

HanabishiRecca:
@bitwiseman, I just think this is not necessary. Because we can't recognize what exactly in the comment section. For people with a such bad style (why just not to move this code to the external js file?) I don't think they care about this.

So I suggest of course detect comments (to avoid `var a = "</script>";` situations) but really ignore insides.

HanabishiRecca:
CDATA is another complicated case. For example I can do
```
alert(/*<![CDATA[*/'<!-- This will be the real js string, not a comment -->'/*]]>*/);
```
изображение


Any random insertions will be syntatically valid
```
var a = 
//<![CDATA[
'<!-- bla -->'
//]]>
;
alert(a);
```
изображение
etc.

So I don't think we want to mess with all this s**t :sweat_smile:

bitwiseman:
This change only effects the one case where you have `<script>` or `<style>` followed immediately by a `<!--` or `<![CDATA[`.   All other cases, including your last examples, are unaffected. 

We already do support beautifying this: 

```
<script>
    <!--
    var a = "Hello, world.";
    -->
</script>
```

I understand that you believe that we should not support this, but that isn't an option.  This change makes the existing feature more robust and makes it work for style tags as well.  Do you have any comments/suggestions on the code?

bitwiseman:
@HanabishiRecca 
On a related topic - Would you like to be involved in triaging features and fixes?  Mostly I've been making those choices and putting items that pass the bar or are higher priority into the current milestone.  If you'd like to be more involved, I'm happy to discuss that, but understand there's a lot of history to consider.

HanabishiRecca:
> This change only effects the one case where you have `<script>` or `<style>` followed immediately by a `<!--` or `<![CDATA[`. All other cases, including your last examples, are unaffected.

Oh, ok then.

> Would you like to be involved in triaging features and fixes?

Unfortunately I don't have enough time to serious dive in development.

bitwiseman:
@HanabishiRecca 
> Unfortunately I don't have enough time to serious dive in development.

Okay, cool.  I want to make sure that you know I am open to input and collaboration in general and that I value your feedback.

PR #1733, opened 2019-11-30
Note that Python version only works with JS files

simonw:
https://stackoverflow.com/questions/41060343/beautify-html-with-js-beautify-from-command-line#comment75350731_41060343

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ To install the Python version of the beautifier:
 ```bash
 $ pip install jsbeautifier
 ```
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,7 +85,7 @@ To install the Python version of the beautifier:
 ```bash
 $ pip install jsbeautifier
 ```
-
+Unlike the JavaScript version, the Python version can only reformat JavaScript. It does not work against HTML or CSS files.
```
bitwiseman: There's a `css-beautify` that also works and I believe is included in the Python package.

PR #1735, opened 2019-12-02
Allow # to start identifier, such as private fields in classes

### LINKED ISSUE #1734: Broken private fields in classes (JS)
# Description
Private fields in JS classes use `#` as part of the identifier name. When beautifying the initial declaration, this library puts it on the wrong line.

# Input
```
class X {
    #__private = null;
}
```

# Expected Output
```
class X {
    #__private = null;
}
```

# Actual Output
```
class X {
    #
    __private = null;
}
```

# Steps to Reproduce

## Environment
OS: Windows
Language: Python

## Settings
Example:
```json
{
    "indent_size": 4,
    "indent_char": " ",
    "indent_level": 0,
    "indent_with_tabs": false,
    "preserve_newlines": true,
    "max_preserve_newlines": 2,
    "jslint_happy": false,
    "space_after_anon_function": false,
    "brace_style": "collapse,preserve-inline",
    "keep_function_indentation": false,
    "eval_code": false,
    "unescape_strings": false,
    "wrap_line_length": 0
}
```

MikeDombo:
To resolve #1734 , this change modifies the tokenizer to recognize #+identifier as a single word. I first tried just adding # as a valid start character, but that broke existing tests, so this is the way I've been able to fix private fields without regressions.

# Description

Fixes Issue: #1734 

# Before Merge Checklist 

Ran `make ci` and all tests are passing including my newly added one.

### DIFF
--- a/js/src/javascript/tokenizer.js
+++ b/js/src/javascript/tokenizer.js
@@ -274,6 +274,12 @@ Tokenizer.prototype._read_non_javascript = function(c) {
         this._input.next();
       }
       return this._create_token(TOKEN.WORD, sharp);
+    } else if (this._input.hasNext()) {
+      // Handle private field (ie. #privateExample)
+      resulting_string = this.__patterns.identifier.read();
+      if (resulting_string) {
+        return this._create_token(TOKEN.WORD, sharp + resulting_string);
+      }
     }
 
     this._input.back();
--- a/js/src/javascript/tokenizer.js
+++ b/js/src/javascript/tokenizer.js
@@ -163,13 +163,13 @@ Tokenizer.prototype._get_next_token = function(previous_token, open_token) { // 
     return this._create_token(TOKEN.EOF, '');
   }
 
+  token = token || this._read_non_javascript(c);
   token = token || this._read_string(c);
   token = token || this._read_word(previous_token);
   token = token || this._read_singles(c);
   token = token || this._read_comment(c);
   token = token || this._read_regexp(c, previous_token);
   token = token || this._read_xml(c, previous_token);
-  token = token || this._read_non_javascript(c);
   token = token || this._read_punctuation();
   token = token || this._create_token(TOKEN.UNKNOWN, this._input.next());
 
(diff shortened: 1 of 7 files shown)


js/src/javascript/tokenizer.js
```diff
@@ -274,6 +274,12 @@ Tokenizer.prototype._read_non_javascript = function(c) {
         this._input.next();
       }
       return this._create_token(TOKEN.WORD, sharp);
+    } else if (this._input.hasNext()) {
+      // Handle private field (ie. #privateExample)
+      resulting_string = this.__patterns.identifier.read();
```
bitwiseman: Maybe we could add `#` to the `__patterns.identifier` as a valid start character?  I understand it isn't valid outside of classes, but it also isn't used for other things right?

MikeDombo: Yes, it was the shebang tests which broke. I'm not sure why they failed since there is specific handling for shebangs which changing the identifier start character shouldn't really have effected.

bitwiseman: The problem is here in the tokenizer: 
https://github.com/beautify-web/js-beautify/blob/c26968ab88761ca1b9cede3844cb52d3d9af03df/js/src/javascript/tokenizer.js#L166-L174

The non-javascript is parsed *after* javascript words.  So, matching for word that start with `#` takes precedence over shebang.  😢 

Please go back and implement this using `__patterns.identifier`.  The try moving the non-javascript detection to before word detection.  Then let's see what errors come up.  Okay?

MikeDombo: Thanks for pointing me in the right direction. Making it evaluate the non-JS stuff first works now with `#` added as a start to an identifier.

test/data/javascript/tests.js
```diff
@@ -169,6 +169,22 @@ exports.test_data = {
           ]
         }
       ]
+    }, {
+      name: "ES5 Class Private Fields",
+      description: "Permit ES5 private class fields which are declared with a leading \"#\".",
+      tests: [
+        { unchanged: '#foo' },
+        {
+          unchanged: [
```
bitwiseman: Please add some tests where the input and output differ.  Then this is will be good to go.

MikeDombo: Added.

MikeDombo:
@bitwiseman, do you know when this might be released? I'd like to use my change of course, but I need it to make it into pypi first.

bitwiseman:
@MikeDombo  
Thanks for the reminder.  Released as 1.10.3 just now.

PR #1746, opened 2020-01-15
Created cssbeautifier PyPI package 

Jorengarenar:
Python version of _css-beautify_ can be installed with `pip`

```sh
$  pip install cssbeautifier
```

_jsbeautifier_ is installed as dependency


bitwiseman:
@Jorengarenar 
This is great, thanks so much! 
I currently using a script to publish all the pieces of this project. 
https://github.com/beautify-web/js-beautify/blob/master/tools/release-all.sh#L20
Will this package get published when I call this? Or do we need to change that script in some way?

Jorengarenar:
@bitwiseman 
Just one line (`python setup-css.py sdist || exit 1`) was everything needed

bitwiseman:
One last question: Would it make sense to install css-beautify as part of the jsbeautifier package?  Rather than having two separate packages?  The node package includes all three.  I'm cool with a separate package, I'm asking your thoughts.

Jorengarenar:
@bitwiseman 
In my personal opinion, when something is not essential, then it shouldn't be installed by default.

Although if it's installed in 99% cases either way, then it should be included in main package and there should be some minimal/headless/base/... version

bitwiseman:
@Jorengarenar 
Thanks!

PR #1752, opened 2020-01-26
Fix required version of jsbeautifier in setup-css.py

Jorengarenar:
`cssbeautifier` won't even install without prior installed `jsbeautifier`, so I've removed unnecessary part about its version in `setup-css.py` and updated README to include this crucial information


bitwiseman:
@Jorengarenar  Hm, windows only error.  Likely infrastructrure issue.

Issue #1776, opened 2020-03-27
Newline rule not working with css-like files


N247S:
# Description
I tried several VSC plugins which handles css formatting (amongst other file types). for bug reports for those I got redirected to the underalying library (which is this project). The basic problem is that I want to have the curly openingbrackets on a new line (after the selector list). But whatever I try, it does not work.

I found an [online beautifier](https://beautifier.io/?without-codemirror), where it didn't work either:

# Input
The code looked like this before beautification:
```
.selector1, .selector2, .selector3 {
	// rules
}
```

# Expected Output
The code should have looked like this after beautification:
```
.selector1,
.selector2,
.selector3
{
	// rules
}
```

# Actual Output
The  code actually looked like this after beautification:
```
.selector1,
.selector2,
.selector3 {
	// rules
}

```

## Settings
```json
{
  "indent_size": "1",
  "indent_char": "\t",
  "max_preserve_newlines": "5",
  "preserve_newlines": true,
  "keep_array_indentation": true,
  "break_chained_methods": false,
  "indent_scripts": "normal",
  "brace_style": "expand",
  "space_before_conditional": true,
  "unescape_strings": true,
  "jslint_happy": false,
  "end_with_newline": false,
  "wrap_line_length": "0",
  "indent_inner_html": true,
  "comma_first": false,
  "e4x": true,
  "indent_empty_lines": true
}
```

Issue #1353, opened 2018-03-02
Does this extension still supports applying Allman style to CSS?


gusbemacbe:
Hello,

Referring to https://github.com/beautify-web/js-beautify/issues/1259. I would like to apply Allman style (expand) to CSS, I hate the KNK style. I applied it and it doesn't work. Observation: The indent size didn't work either. 

## Environment
**OS verison:** macOS 10.13.3
**VSCode version:** Insider 1.21.0

## Settings
```json
"beautify.options":
{
        "brace_style": "expand",
        "indent_size": 2,
        "css": 
        {
            "brace_style": "expand",
            "indent_size": 1
        },
}
```

bitwiseman:
This is not currently supported.  I've added this to the backlog.

gusbemacbe:
I am waiting yet...

bitwiseman:
@proimage 
As with so many features of the project, this is awaiting someone to take ownership of it and drive it to completion.  Looking at it, it is probably something that someone could add with relative ease.  You interested?

gusbemacbe:
Hi @bitwiseman !

I am ready to be a collaborator as I got graduated in computer science and received a diploma yesterday. But I am not totally advanced in JavaScript and in Nodejs. I am studying them in the moment. May I take?

proimage:
If I knew how to program, I might. Sorry!

bitwiseman:
@gusbemacbe 
Go right ahead.  You don't need to be advanced.  
The change will be in this area: https://github.com/beautify-web/js-beautify/blob/master/js/src/css/beautifier.js#L286-L310

Come over to the project gitter channel if you want to chat.

gusbemacbe:
Better to fork and git-clone this repository and test it manually before to pull request. I have to run `node run` inside the git-cloned folder.

gusbemacbe:
It seems to give a small error.

bitwiseman:
@gusbemacbe 
We use `make` to run build and test.  
See https://github.com/beautify-web/js-beautify/blob/master/CONTRIBUTING.md

Issue #1259, opened 2017-09-25
Add brace_style option for CSS


Hirse:
# Description
Based on this Stackoverflow [question](https://stackoverflow.com/q/45708124/1469028) and [answer](https://stackoverflow.com/a/46304248/1469028) there seems to be a desire to configure the brace style for CSS in order to have opening braces on a new line.

# Input
The code looked like this before beautification:
```css
#icon {
    background-color: blue;
}
```

# Expected Output
The code should have looked like this after beautification:
```css
#icon
{
    background-color: blue;
}
```

# Actual Output
The  code actually looked like this after beautification:
```css
#icon {
    background-color: blue;
}
```

## Settings
```json
{
    "brace_style": "expand"
}
```

matt-walker-aib:
Oh, this would make my life sooo much better...

PR #1796, opened 2020-05-08
Add expand brace-style option to css beautifier

ang-zeyu:
# Description

Fixes Issue: #1776 ( adding a newline before the opening curly brace of selector(s) )
#1353 

#1259 (partially, till the other options are implemented)

~Name of the new option: `curly-start-newline`~
~Shorthand: `-R` for the `cur` in `curly`~

Added `brace-style=expand` to the css beautifier

# Before Merge Checklist 

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,7 @@ CSS Beautifier Options:
   -t, --indent-with-tabs             Indent with tabs, overrides -s and -c
   -e, --eol                          Character(s) to use as line terminators. (default newline - "\\n")
   -n, --end-with-newline             End output with newline
+  -b, --brace-style                  [collapse|expand] ["collapse"]
   -L, --selector-separator-newline   Add a newline between multiple selectors
   -N, --newline-between-rules        Add a newline between CSS rules
   --indent-empty-lines               Keep indentation on empty lines
--- a/js/src/css/beautifier.js
+++ b/js/src/css/beautifier.js
@@ -291,23 +291,34 @@ Beautifier.prototype.beautify = function() {
         insidePropertyValue = false;
         this.outdent();
       }
-      this.indent();
-      this._output.space_before_token = true;
-      this.print_string(this._ch);
 
       // when entering conditional groups, only rulesets are allowed
       if (enteringConditionalGroup) {
         enteringConditionalGroup = false;
-        insideRule = (this._indentLevel > this._nestedLevel);
+        insideRule = (this._indentLevel >= this._nestedLevel);
       } else {
         // otherwise, declarations are also allowed
-        insideRule = (this._indentLevel >= this._nestedLevel);
+        insideRule = (this._indentLevel >= this._nestedLevel - 1);
       }
       if (this._options.newline_between_rules && insideRule) {
         if (this._output.previous_line && this._output.previous_line.item(-1) !== '{') {
           this._output.ensure_empty_line_above('/', ',');
         }
       }
+
+      this._output.space_before_token = true;
+
+      // The difference in print_string and indent order is necessary to indent the '{' correctly
+      if (this._options.brace_style === 'expand') {
+        this._output.add_new_line();
+        this.print_string(this._ch);
+        this.indent();
+        this._output.set_indent(this._indentLevel);
+      } else {
+        this.indent();
+        this.print_string(this._ch);
+      }
+
       this.eatWhitespace(true);
       this._output.add_new_line();
     } else if (this._ch === '}') {
--- a/js/src/css/options.js
+++ b/js/src/css/options.js
@@ -38,6 +38,16 @@ function Options(options) {
   var space_around_selector_separator = this._get_boolean('space_around_selector_separator');
   this.space_around_combinator = this._get_boolean('space_around_combinator') || space_around_selector_separator;
 
+  var brace_style_split = this._get_selection_list('brace_style', ['collapse', 'expand', 'end-expand', 'none', 'preserve-inline']);
+  this.brace_style = 'collapse';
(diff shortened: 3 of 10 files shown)


js/src/css/beautifier.js
```diff
@@ -291,23 +291,34 @@ Beautifier.prototype.beautify = function() {
...
         // otherwise, declarations are also allowed
-        insideRule = (this._indentLevel >= this._nestedLevel);
+        insideRule = (this._indentLevel >= this._nestedLevel - 1);
       }
       if (this._options.newline_between_rules && insideRule) {
         if (this._output.previous_line && this._output.previous_line.item(-1) !== '{') {
           this._output.ensure_empty_line_above('/', ',');
         }
       }
+
```
ang-zeyu, line 308: moved this section below to prevent ( when using `newline_between_rules` )
`.a{} .b{}` -->
```
.a
/* there shouldn't be an empty line here */
{}

.b
/* there shouldn't be an empty line here */
{}
```

On the other hand the branching is to indent the `{` correctly

README.md
```diff
@@ -323,6 +323,7 @@ CSS Beautifier Options:
   -e, --eol                          Character(s) to use as line terminators. (default newline - "\\n")
   -n, --end-with-newline             End output with newline
   -L, --selector-separator-newline   Add a newline between multiple selectors
+  -C, --curly-start-newline          Add a newline before the opening { bracket of selector(s)
```
bitwiseman, line 326: What do you think of reusing the `brace-style` setting instead of create a new field specific to the css beautifier? 
I ask because the formatting `curly-start-newline` creates is the same as `brace-style=expand`.  The default behavior matches `brace-style=collapse`  (or `end-expand` but these are the same in css.) 

Don't change anything yet, I'd like to discuss and see if what I'm thinking makes sense to you.

ang-zeyu: Thanks for looking at this!

Think it would be more consistent to use `brace-style` as well (I missed it out before).
Perhaps a trimmed set of options like `[collapse|expand|none]`, since as you mentioned `end-expand` dosen't make sense in css. I'm not too sure if `preserve-inline` makes sense though. (adding css to a minified file while preserving the minified styles maybe?)

I could expand the scope of this to include support for these if needed, or in separate PRs. Let me know what you think!

bitwiseman: @ang-zeyu
I think it would be fine to start with support for just `collapse` and `expand` (and treat `end-expand` and `none` the same as `collapse`).  Then you could open an issue to add support for `preserve-inline` and `none` and someone (or you) can do that at a later date in another PR. 

That will get this cool new feature out for people to use with minimal added work for you now.  Sound good?

ang-zeyu: Yup! will update this with `brace-style=[collapse|expand]` as a start then

ang-zeyu:
ready for another look @bitwiseman

I noticed `braces_on_own_line` is still supported in the javascript options though, should that be added here as well?

bitwiseman:
@ang-zeyu 
re `braces_on_own_line`: no.  You're implementing something new, backward compatibility is not your concern.  If someone wants to use your feature, it is fine to require them to use current settings.

test/data/css/tests.js
```diff
@@ -340,27 +351,79 @@ exports.test_data = {
...
+          ]
         },
-        { input: '#bla, #foo{color:black}', output: '#bla,{{separator}}#foo {\n    color: black\n}' }, {
+        {
+          input: '#bla, #foo{color:black}',
+          output: '#bla,{{separator}}#foo{{curly_separator}}{\n    color: black\n}'
+        },
+        {
           input: 'a:first-child,a:first-child{color:red;div:first-child,div:hover{color:black;}}\na:first-child,a:first-child{color:red;div:first-child,div:hover{color:black;}}',
-          output: 'a:first-child,{{separator}}a:first-child {\n    color: red;{{new_rule}}\n    div:first-child,{{separator1}}div:hover {\n        color: black;\n    }\n}\n{{new_rule}}a:first-child,{{separator}}a:first-child {\n    color: red;{{new_rule}}\n    div:first-child,{{separator1}}div:hover {\n        color: black;\n    }\n}'
```
bitwiseman, line 363: Excellent, thanks for cleaning up this formatting.

js/src/css/options.js
```diff
@@ -38,6 +38,11 @@ function Options(options) {
   var space_around_selector_separator = this._get_boolean('space_around_selector_separator');
   this.space_around_combinator = this._get_boolean('space_around_combinator') || space_around_selector_separator;
 
+  var brace_style_split = this._get_selection_list('brace_style', ['collapse', 'expand']);
```
bitwiseman, line 41: You'll need to accept the other values that are allowed by the javascript options, otherwise you'll get an error if someone passes them (via inheriting the option). You can ignore or convert them loop below.

ang-zeyu: 👍 done. Linked two more related issues as well

js/src/css/options.js
```diff
@@ -38,6 +38,16 @@ function Options(options) {
   var space_around_selector_separator = this._get_boolean('space_around_selector_separator');
   this.space_around_combinator = this._get_boolean('space_around_combinator') || space_around_selector_separator;
 
+  var brace_style_split = this._get_selection_list('brace_style', ['collapse', 'expand', 'end-expand', 'none', 'preserve-inline']);
+  this.brace_style = 'collapse';
+  for (var bs = 0; bs < brace_style_split.length; bs++) {
+    if (brace_style_split[bs] !== 'expand') {
+      // default to collapse, as only collapse|expand is implemented for now
```
bitwiseman, line 45: Perfect.

Issue #1799, opened 2020-05-17
 selector_separator_newline adds erroneous newline on @extend SCSS statements


TomAnthony:
# Description

The `selector_separator_newline` option, when set to `true`, is slightly buggy with SCSS `@extend` statements, adding a newline when it should not.

# Input
The code looked like this before beautification:

```scss
	a {
		@extend .color-black, .hover-far-red;
	}
```

# Expected Output
The code should have looked like this after beautification:

```scss
	a {
		@extend .color-black, .hover-far-red;
	}
```

# Actual Output
The  code actually looked like this after beautification:

```scss
	a {
		@extend .color-black,
		.hover-far-red;
	}
```

The added newline is erroneous.

It works correctly for `@import` statements with comma delimited entries, and does not add a new line.

# Steps to Reproduce

## Settings

Example:
```json
{
    "selector_separator_newline": false
}
```

damienfa:
Same issue everywhere there is a "comma separated list" . 

Example with list variable definition in SCSS :

**Expected :** 
`$default-margin: (desktop: 24px, tablet: 16px, phone: 16px);`

**Actual :** 
```
$layout-grid-default-margin: (desktop: 24px,
tablet: 16px,
phone: 16px,
);
```

mhnaeem:
@bitwiseman This was fixed in #1800

PR #1800, opened 2020-05-17
Fix erroneous newline added on @extend statements

TomAnthony:
# Description

Fixes Issue: https://github.com/beautify-web/js-beautify/issues/1799

I'm unsure if this needs new tests?

# Before Merge Checklist 

- *NA* ~Added command-line option(s) (NA if~
- *NA* ~README.md documents new feature/option(s)~


bitwiseman:
@TomAnthony 
Yes, this needs new tests. The old ones didn't catch this issue so we need tests to cover it.

bitwiseman:
@TomAnthony  Do you plan to write tests for this?

TomAnthony:
@bitwiseman - it is on my list to do, but I'm not familiar with the test setup so it would need a bit of time, which I'm short on at the moment.

If someone else wants to dive in to write some tests that would be great, otherwise I will try to get to it.

TomAnthony:
@bitwiseman - I have added a test for this now. 

 I'm not v. familiar with the test setup, so please check, but I can confirm it fails on the current `main` branch, but passes on this branch.
 
 Thanks again for the plugin! :)

bitwiseman (approved):

Issue #1718, opened 2019-10-13
no new line after self closing tag


Steve-Mcl:
# Description
no new line after self closing tag.
I have searched closed and open issues but couldnt find any issue identical to this one.

# Input
The code looked like this before beautification:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="none" x="0" y="0" viewBox="0 0 900 710" width="100%" height="100%">
  <circle id="mycircle" 
  cx="182.901" cy="91.4841" 
  style="fill:rosybrown;stroke:black;stroke-width:1px;" r="48" /></svg>
```

# Expected Output
The code should have looked like this after beautification:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="none" x="0" y="0" viewBox="0 0 900 710" width="100%" height="100%">
  <circle id="mycircle" cx="182.901" cy="91.4841" style="fill:rosybrown;stroke:black;stroke-width:1px;" r="48" />
</svg>
```

# Actual Output
The  code actually looked like this after beautification:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="none" x="0" y="0" viewBox="0 0 900 710" width="100%" height="100%">
  <circle id="mycircle" cx="182.901" cy="91.4841" style="fill:rosybrown;stroke:black;stroke-width:1px;" r="48" /></svg>
```

# Steps to Reproduce
In chrome (on windows) using the above original HTML/SVG and the below settings.

## Environment
OS: Windows / Chrome

## Settings
Example:
```json
    {
                    "indent_size": "2",
                    "indent_char": " ",
                    "max_preserve_newlines": "-1",
                    "preserve_newlines": false,
                    "keep_array_indentation": false,
                    "break_chained_methods": true,
                    "indent_scripts": "normal",
                    "brace_style": "expand",
                    "space_before_conditional": true,
                    "unescape_strings": false,
                    "jslint_happy": false,
                    "end_with_newline": false,
                    "wrap_line_length": "0",
                    "indent_inner_html": false,
                    "comma_first": false,
                    "e4x": false,
[...]

Issue #1365, opened 2018-04-05
HTML format, no break after <label>?


hstig:
# Input
The code looked like this before beautification:
```
<div class="col-xs-2">
  <input type="radio" class="control-label" ng-disabled="!col" ng-model="col" value="2" class="form-control" id="coli" name="coli" />
  <label for="coli" class="control-label">Collision</label></div>


```

# Expected Output
The code should have looked like this after beautification:
```
<div class="col-xs-2">
  <input type="radio" class="control-label" ng-disabled="!col" ng-model="col" value="2" class="form-control" id="coli" name="coli" />
  <label for="coli" class="control-label">Collision</label>
</div>
```

# Actual Output
The  code actually looked like this after beautification:
```
<div class="col-xs-2">
  <input type="radio" class="control-label" ng-disabled="!col" ng-model="col" value="2" class="form-control" id="coli" name="coli" />
  <label for="coli" class="control-label">Colision</label></div>
```

# Steps to Reproduce

js-beautify --html -r -s 2 -b collapse 
## Environment
OS: debian etch
js-beautify 1.7.5

```

gabrielhfgomes:
Is this bug solved. I got the same formatting problem!

bitwiseman:
@gabrielhfgomes 
Nope, still there.  Help welcome.

gabrielhfgomes:
I was having this problem with beautify in VSCode, I solved it now by rolling back the extension to a 2 year behind version, the 1.0.2 and than the formatting worked perfectly as expected.

bitwiseman:
@gabrielhfgomes 
Not impossible, the code from two years ago was completely different and probably didn't even test properly.  Thanks for trying.

bitwiseman:
This line will cause tags like that `</div>` to be moved to a new line:
https://github.com/beautify-web/js-beautify/blob/master/js/src/html/beautifier.js#L706

However blocks like this are what set that flag: 
https://github.com/beautify-web/js-beautify/blob/333710642f1fdf1fddd65ab8e7fe9ace483ca4c5/js/src/html/beautifier.js#L728-L733

That is obviously not getting set in some key scenarios.

PR #1819, opened 2020-07-19
Fix missing line break at end tag

ang-zeyu:
# Description

Fixes Issue: #1718, #1365

# Before Merge Checklist 

- [x] Python implementation (NA)
- [x] Added command-line option(s) (NA)
- [x] README.md documents new feature/option(s) (NA)

### DIFF
--- a/index.html
+++ b/index.html
@@ -206,7 +206,8 @@
         <li>A
           <a href="javascript:(function(){s=document.getElementsByTagName('SCRIPT');tx='';sr=[];for(i=0;i<s.length;i++){with(s.item(i)){t=text;if(t){tx+=t;}else{sr.push(src)};}};with(window.open()){document.write('<textarea id=&quot;t&quot;>'+(sr.join(&quot;\n&quot;))+&quot;\n\n-----\n\n&quot;+tx+'</textarea><script src=&quot;https://beautifier.io/js/lib/beautify.js&quot;></script><script>with(document.getElementById(&quot;t&quot;)){value=js_beautify(value);with(style){width=&quot;99%&quot;;height=&quot;99%&quot;;borderStyle=&quot;none&quot;;}};</script>');document.close();}})();">
             <strong>bookmarklet</strong></a>
-          (drag it to your bookmarks) by Ichiro Hiroshi to see all scripts used on the page,</li>
+          (drag it to your bookmarks) by Ichiro Hiroshi to see all scripts used on the page,
+        </li>
 
         <li><strong>Chrome</strong>, in case the built-in CSS and javascript formatting isn't enough for you:<br>
           — <a href="https://chrome.google.com/webstore/detail/cfmcghennfbpmhemnnfjhkdmnbidpanb">Quick source viewer</a> by Tomi Mickelsson (<a href="https://github.com/tomimick/chrome-ext-view-src">github</a>, <a href="http://tomicloud.com/2012/07/viewsrc-chrome-ext">blog</a>),<br>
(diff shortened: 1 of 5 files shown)


ang-zeyu:
#### Details

there's an explicit check against this in `set_tag_position` for inline elements (`svg`), it would work correctly without this PR if the `svg` in #1718 was switched out for a `div`

Current code:

```
} else if (parser_token.is_end_tag) { //this tag is a double tag so check for tag-ending
    /*
     if (either):
     1. start tag is multiline_content
     2. not inline el and
        previous tag token not inline and
        (previous tok not TAG_CLOSE or (ie. has some intermediate stuff in between)
         start_tag_token !== last_tag_token (ie. has some intermediate tags in between)) and
        previous tok is not text
     print a newline
     */
    if ((parser_token.start_tag_token && parser_token.start_tag_token.multiline_content) ||
      !(parser_token.is_inline_element || // this is causing the problem
        (last_tag_token.is_inline_element) ||
        (last_token.type === TOKEN.TAG_CLOSE &&
          parser_token.start_tag_token === last_tag_token) ||
        (last_token.type === 'TK_CONTENT')
      )) {
      printer.print_newline(false);
    }
  }
```

Originally introduced in https://github.com/beautify-web/js-beautify/commit/0e9ad61 (#1407)

It's an important check though, newlines shouldn't be printed before inline elements.

#### Fix

Currently `multiline_content` (correct me if im wrong) is set on the basis that some child of it is a 'block' (non-inline, non-text) element for _start tags_. This fails for the example in the issue as it's an 
'empty element'.

This pr just sets `multiline_content` correctly for the above case.

ang-zeyu:
Expanded the scope of this pr to include (#1365) since the symptoms are similar;

### Second commit (#1365)

As mentioned here https://github.com/beautify-web/js-beautify/issues/1365#issuecomment-496613290, this commit adds newline detection for starting _inline_ tags. If the parent is a non-inline element, we set `multiline_content` to true here as well.

I'm not sure of the expected behaviour for ending inline tag tokens with newlines though (left as is):

```
<div class="col-xs-2"><label for="coli" class="control-label">Collision
</label></div>
```

currently beautifies into:
```
<div class="col-xs-2"><label for="coli" class="control-label">Collision
    </label></div>
```

if we add a similar check,

```
<div class="col-xs-2"><label for="coli" class="control-label">Collision
    </label>
</div>
```

which should be the correct behaviour?

bitwiseman:
@ang-zeyu 

To answer your question, while the second output feels correct, I think the first output is actually correct.  The new line occurs inside the `<label>` meaning it is multiline, but that doesn't not make the parent `<div>` multiline:

```
<div class="col-xs-2"><label for="coli" class="control-label">Collision
    </label></div>
```

However, I did find a different case that is not correct. 

Input
```
<div class="col-xs-2">
    <input type="radio" class="control-label" ng-disabled="!col" ng-model="col" value="2" class="form-control" id="coli" name="coli" /></div>
```

Current output is same as above. 

Expected output: 
```
<div class="col-xs-2">
    <input type="radio" class="control-label" ng-disabled="!col" ng-model="col" value="2" class="form-control" id="coli" name="coli" />
</div>
```

Don't worry about fixing this issue, I'll make some tweaks to what you did and merge.  Thanks!

ang-zeyu:
> To answer your question, while the second output feels correct, I think the first output is actually correct. The new line occurs inside the <label> meaning it is multiline, but that doesn't not make the parent <div> multiline:

agreed, that makes sense.

> Don't worry about fixing this issue, I'll make some tweaks to what you did and merge. Thanks!

neat, thanks for looking at this as well!

bitwiseman:
This is turned into a significantly more complex change.  Adding tests for edge cases exposed more issues (most of the preexisting) that needed fixing.  

The result is an implementation with fewer code paths, but more churn in this PR.

PR #1833, opened 2020-08-22
Use native mkdirSync instead of 'mkdirp' package

Krinkle:
# Description

In Node.js 10 and later, recursive mkdir is available in core. No longer needs the `mkdirp` package.
<https://nodejs.org/docs/latest-v10.x/api/fs.html#fs_fs_mkdirsync_path_options>

# Before Merge Checklist 

- [x] Added command-line option(s)

Ref https://github.com/11ty/eleventy/issues/1375.

### DIFF
--- a/js/src/cli.js
+++ b/js/src/cli.js
@@ -40,7 +40,6 @@ var debug = process.env.DEBUG_JSBEAUTIFY || process.env.JSBEAUTIFY_DEBUG ? funct
 var fs = require('fs'),
     cc = require('config-chain'),
     beautify = require('../index'),
-    mkdirp = require('mkdirp'),
     nopt = require('nopt'),
     glob = require('glob');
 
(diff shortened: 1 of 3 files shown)


Krinkle:
@bitwiseman The main motivation for this is to reduce the size of the dependency graph for downstream users of `js-beautify`, thus requiring fewer packages to be audited and trusted.

As Node 8 is beyond its EOL date, I assumed requiring Node 10 for new releases would be uncontroversial (<https://github.com/nodejs/Release>). I'll see about making those PRs!

bitwiseman:
@Krinkle 
You'd be surprised about what is controversial for this library.  
I totally agree with reducing the dependency graph.  We just tend to move really slowly on breaking changes like this.

Krinkle:
> You'd be surprised about what is controversial for this library.

To clarify, I didn't mean that it _should_ be uncontroversial, just that I figured there's a good chance that it is given it was no longer tested (and EOL). If correct, that this change can help improve something downstream. But, if support is meant to remain, that's totally fine with me! I'm just going through the large tree I'm currently auditing, simplifying code where possible and sometimes proposing to remove potentially obsolete fallbacks.

jimmywarting (approved):

Krinkle:
Merge conflict resolved.

PR #1841, opened 2020-09-28
fix bug in node require

scil:
Fixes Issue: 
`var beautify = require('js-beautify/').js` not work, because js-beautifify index.js is located in `node_modules/js-beautify/js/index.js`.

So 'js-beautify/js` is needed.


VillainsRule4000 (approved):
simple, but works well

Issue #1856, opened 2020-11-18
Use the space_in_paren option in CSS files too.


ninio:
# Description
At the moment the `space_in_paren` option is used only in the JS files and there is no way to enable it in the CSS beautification.

# Input
With this new feature, when I give like this input: 
```
.test {
  height: min(100%, 100vh);
}
```

# Expected Output
I'd like to see this output:
```
.test {
  height: min( 100%, 100vh );
}
```

PR #1857, opened 2020-11-19
Added space_in_paren to the CSS options

ninio:
The default value is false to keep the current behaviour as-is
Added both JS and Py implementations
Added unit tests for the space_in_paren option

# Description

Fixes Issue: 
#1856

# Before Merge Checklist 

- [ ] Added command-line option(s) - It's already documented, but was not applied for CSS beautification.
- [ ] README.md documents new feature/option(s) - It's already documented, but was not applied for CSS beautification.

### DIFF
--- a/test/data/css/tests.js
+++ b/test/data/css/tests.js
@@ -1800,6 +1800,48 @@ exports.test_data = {
         ]
       }]
     }, {
+      name: "space_in_paren true",
+      description: "",
+      options: [
+        { name: "space_in_paren", value: "true" }
+      ],
+      tests: [{
+        fragment: true,
+        input: [
+          'a {',
+          'width: min(100%,100vw);',
+          'height: calc( 100vh - 100px );',
+          '}'
+        ],
+        output: [
+          'a {',
+          '    width: min( 100%, 100vw );',
+          '    height: calc( 100vh - 100px );',
+          '}'
+        ]
+      }]
+    }, {
+      name: "space_in_paren false",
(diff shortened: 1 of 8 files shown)


test/data/css/tests.js
```diff
@@ -1800,6 +1800,48 @@ exports.test_data = {
...
+      name: "space_in_paren true",
+      description: "",
+      options: [
+        { name: "space_in_paren", value: "true" }
+      ],
+      tests: [{
+        fragment: true,
+        input: [
+          'a {',
+          'width: min(100%,100vw);',
```
bitwiseman, line 1812: Please add another test where the input has multiple spaces in inside the paren, example: `'width: min(. 
  100%. ,  100vw.   );',`.  In the case the extra spaces should be removed.

bitwiseman (changes requested):
Thanks for contributing!

Additional testing needed. 

You might also try using a `matrix` config for the test.   That way you can can supply a few inputs and generate the expected outputs like in the example below.  

https://github.com/beautify-web/js-beautify/blob/3f1fff2b0ab58833806e9487de9b89945a371e46/test/data/css/tests.js#L1670-L1714

test/data/css/tests.js
ninio: Thanks for the feedback! Will add them in the coming week.

bitwiseman:
@ninio  Do you have time to update this?

Issue #1244, opened 2017-09-11
Tagged Template literals


laggingreflex:
# Description

Are there any rules that apply to [Tagged Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals)? 

Specifically:

* # Case 1:

    ## Input
    ```
    fn`tagged`
    ```
    ## Expected: no change
    ```
    fn`tagged`
    ```
    ## Actual: space inserted
    ```
    fn `tagged`
      ^ space inserted
    ```

* # Case 2:

    ## Input
    ```
    fn()`tagged`
    ```
    ## Expected: no change
    ```
    fn()`tagged`
    ```
    ## Actual: newline inserted
    ```
    fn()
    `tagged`
    ^ newline inserted
    ```

bitwiseman:
Not currently supported.  Didn't even know this existed. Thanks!

JuanFML:
Hello everyone, I am new here, is there a chance I can try this as my first issue?

bitwiseman:
@JuanFML 
You are welcome to give it a try.  You'll probably want to make the change in this method: 

https://github.com/beautify-web/js-beautify/blob/97caa308491e1379b226c514f677fa4f278e15d4/js/src/javascript/beautifier.js#L1059-L1077

If the `STRING` token that is about to be processed is a backtick string and  is preceded by a literal with no space in between, then do not add a space or newline.

JuanFML:
Ok! I will be trying it out, thank you! If I get stuck I will let you know

JuanFML:
Hi @bitwiseman, I've been working on this, but I would like to know if I am doing it right, so far I figure out a condition and added it in the method you mentioned, like so: 
```
Beautifier.prototype.handle_string = function(current_token) {

  if (/^[`][^`]+[`]$/gm.test(current_token.text) && (this._flags.last_token.type === 'TK_END_EXPR' || this._flags.last_token.type === 'TK_WORD') && current_token.whitespace_before === '') {
  } else if (this.start_of_statement(current_token)) {
    // The conditional starts the statement if appropriate.
    // One difference - strings want at least a space before
    this._output.space_before_token = true;
  } else {
    this.handle_whitespace_and_comments(current_token);
    if (this._flags.last_token.type === TOKEN.RESERVED || this._flags.last_token.type === TOKEN.WORD || this._flags.inline_frame) {
      this._output.space_before_token = true;
[...]

PR #1860, opened 2020-12-03
Fixing bug #1244

JuanFML:
Added validations to prevent tagged literals from being modified

# Description

Fixes Issue:  #1244 

# Before Merge Checklist 

### DIFF
--- a/js/src/javascript/beautifier.js
+++ b/js/src/javascript/beautifier.js
@@ -1066,15 +1066,15 @@ Beautifier.prototype.handle_string = function(current_token) {
     }
   } else {
     this.handle_whitespace_and_comments(current_token);
-    if (this._flags.last_token.type === TOKEN.RESERVED || this._flags.last_token.type === TOKEN.WORD || this._flags.inline_frame) {
-      if (!(backtickRegex.test(current_token.text) && (this._flags.last_token.type === TOKEN.END_EXPR || this._flags.last_token.type === TOKEN.WORD) && current_token.whitespace_before === '')) {
-        this._output.space_before_token = true;
-      }
+    if ((backtickRegex.test(current_token.text) && (this._flags.last_token.type === TOKEN.END_EXPR || this._flags.last_token.type === TOKEN.WORD) && current_token.whitespace_before === '')) {
+      //This conditionial checks backtick strings and makes no changes
+    } else if (this._flags.last_token.type === TOKEN.RESERVED || this._flags.last_token.type === TOKEN.WORD || this._flags.inline_frame) {
+      this._output.space_before_token = true;
     } else if (this._flags.last_token.type === TOKEN.COMMA || this._flags.last_token.type === TOKEN.START_EXPR || this._flags.last_token.type === TOKEN.EQUALS || this._flags.last_token.type === TOKEN.OPERATOR) {
       if (!this.start_of_object_property()) {
         this.allow_wrap_or_preserved_newline(current_token);
       }
-    } else if ((backtickRegex.test(current_token.text) && (this._flags.last_token.type === TOKEN.END_EXPR || this._flags.last_token.type === TOKEN.WORD) && current_token.whitespace_before === '')) {} else {
+    } else {
       this.print_newline();
     }
   }
--- a/js/src/javascript/beautifier.js
+++ b/js/src/javascript/beautifier.js
@@ -1057,7 +1057,7 @@ Beautifier.prototype.handle_semicolon = function(current_token) {
 };
 
 Beautifier.prototype.handle_string = function(current_token) {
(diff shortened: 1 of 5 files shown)


js/src/javascript/beautifier.js
```diff
@@ -1057,7 +1057,10 @@ Beautifier.prototype.handle_semicolon = function(current_token) {
 };
 
 Beautifier.prototype.handle_string = function(current_token) {
-  if (this.start_of_statement(current_token)) {
+  var backtickRegex = /^[`].+[`]$/gm;
+  if ((backtickRegex.test(current_token.text) && (this._flags.last_token.type === TOKEN.END_EXPR || this._flags.last_token.type === TOKEN.WORD) && current_token.whitespace_before === '')) {
```
bitwiseman, line 1061: Instead of regex-ing for the whole text, would this work? 
```suggestion
  if (current_token.text.startsWith("`") && (this._flags.last_token.type === TOKEN.END_EXPR || this._flags.last_token.type === TOKEN.WORD) && current_token.whitespace_before === '') {
```

python/jsbeautifier/javascript/beautifier.py
```diff
@@ -1184,13 +1184,21 @@ def handle_semicolon(self, current_token):
         self.print_token(current_token)
 
     def handle_string(self, current_token):
-        if self.start_of_statement(current_token):
+        if (
+            current_token.text[0]
+            == current_token.text[len(current_token.text) - 1]
+            == "`"
```
bitwiseman, line 1190: I see here you skipped the regex.
```suggestion
            current_token.text[0]
            == "`"
```

bitwiseman (changes requested):
Wow, you basically got it. Great contribution!
Minor suggestions/changes. 

Also, could you add tests for inputs such as the following (some of which are expected change):
```javascript
fn() `tagged`

// fn()\n`tagged`
fn()
`tagged`

`untagged`+`untagged`

// this is invalid, right? I think a space should be added. 
fn[0]`tagged`
```

JuanFML:
Thank you! I will make changes you suggested, but I just have a question for the last two tests, in the second one, should the output be the same?: 

```unchanged: `untagged`+`untagged` ```

And fo the last one, I am not sure but I think it would be invalid as you said, so ashould look like this:
```
input: fn[0]`tagged`
output: fn[0] `tagged`
```

**Comment by JuanFML** [comment]
I am having a little trouble with the first test you mentioned, I have added the test like so:

```
{
 input: "fun() `tagged`",
 output: "fun()\n`tagged`"
},
```

I am getting the next error:
```
==== more random test ============================================================
---- input -------
fun()
`tagged`
---- expected ----
fun()
`tagged`
---- output ------
fun()`tagged`
---- expected-ws ------
fun()\n
`tagged`
---- output-ws ------
fun()`tagged`
================================================================

```

**Comment by bitwiseman** [comment]
@JuanFML 
Ah, `whitespace` is the string of whitespace chars on the current line before the current token.  Newlines are counted up separately as `newlines`. 

https://github.com/beautify-web/js-beautify/blob/97caa308491e1379b226c514f677fa4f278e15d4/js/src/core/token.js#L43-L44

**Comment by bitwiseman** [comment]
> Thank you! I will make changes you suggested, but I just have a question for the last two tests, in the second one, should the output be the same?:
> 
> ```
> input: `untagged`+`untagged` 
> output:  `untagged` + `untagged` 
> ```

This is correct. 

> 
> And fo the last one, I am not sure but I think it would be invalid as you said, so ashould look like this:
> 
> ```
> input: fn[0]`tagged`
> output: fn[0] `tagged` 

Please do this one. ^

> or
> output: fn[0]
> `tagged`
> ```

**Comment by JuanFML** [comment]
I think I have fixed it, let me know if it is okay this way!

File: test/data/javascript/tests.js [hunk 3]
```diff
@@ -1708,6 +1708,30 @@ exports.test_data = {
...
+        {
+          unchanged: 'fn`${fn2()} more text ${`${`more text`}`} banana ${fn3`test`} ${fn4()`moretest banana2`}`'
+        },
+        {
+          input: '`untagged`+`untagged`',
+          output: '`untagged` + `untagged`'
+        },
+        {
+          input: "fun() `taggedd`",
+          output: "fun()\n`taggedd`"
```
bitwiseman, line 1730:
```suggestion
          output: "fun() `taggedd`"
```

JuanFML: So this should be unchanged then?

bitwiseman: @JuanFML
Unchanged.  If they put a space in there, it isn't valid syntax but this tool should not assume it knows what they meant.

JuanFML: I have added the changes

bitwiseman (approved):

Issue #1854, opened 2020-11-02
React Fragment Short Syntax <></> issue


laggingreflex:
# Description

[React Fragment Short Syntax][1] gets messed up.

The beautified code results in compile errors. 

# Input
```
class Columns extends React.Component {
  render() {
    return (
      <>        
        <td>Hello</td>
        <td>World</td>
      </>    
    );
  }
}
```

# Expected Output

Shouldn't change:

```
class Columns extends React.Component {
  render() {
    return (
      <>        
        <td>Hello</td>
        <td>World</td>
      </>    
    );
  }
}
```

# Actual Output
The  code actually looked like this after beautification:
```
class Columns extends React.Component {
    render() {
        return (
            <>        
        <td>Hello</td> <
            td > World < /td> <
            />    
        );
    }
}
```

# Steps to Reproduce

Copy paste the [the sample code from docs][1] on https://beautifier.io/

## Environment

Tried on https://beautifier.io/

## Settings

Default on https://beautifier.io/ (with jsx)

[1]: https://reactjs.org/docs/fragments.html#short-syntax

bitwiseman:
Need to update the `xml` tokenizer to capture an empty tag. 

https://github.com/beautify-web/js-beautify/blob/97caa308491e1379b226c514f677fa4f278e15d4/js/src/javascript/tokenizer.js#L129

JuanFML:
Hello! I was looking into this issue, could I work on it?

Issue #665, opened 2015-04-08
Beautifier does not re-indent E4X/JSX literals


bitwiseman:
```js
// This remains badly indented
var a = <x>
<note>everything {inside} the xml tags is treated as a literal string. /sadface</note>
</x>
```

zebzhao:
I wrote an indentation library that handles jsx. If anyone still needs jsx indentation, you can have a look. https://github.com/zebzhao/indent.js

bitwiseman:
@zebzhao 
This is not the place to advertise your project, sorry.
If you'd like to contribute code to fix this issue in this project, we'd welcome the help.

zebzhao:
Then can you please fix this? I really need this.

bitwiseman:
As you're probably noticing from the issues filed on your project indenting
well is harder than it first appears.
I do this project in my spare time.  That is why I said you contribution
would be helpful.
I'll see what I can do.

bitwiseman:
It might be possible to look at the incoming code to try to detect what the indenting was and then use that to replace the leading whitespace for each line in the xml.

Long term the right thing would be to extract xml and/or react beautifiers from the html beautifier and have the javascript beautifier call out to that when it encounters an xml block inside js.

PR #1868, opened 2020-12-23
Capture empty tags

JuanFML:
# Description
Updated xml to capture empty tags in both python and JS files

Fixes Issue:  #1854 

# Before Merge Checklist 

### DIFF
--- a/js/src/javascript/tokenizer.js
+++ b/js/src/javascript/tokenizer.js
@@ -126,7 +126,7 @@ var Tokenizer = function(input_string, options) {
     html_comment_end: pattern_reader.matching(/-->/),
     include: pattern_reader.starting_with(/#include/).until_after(acorn.lineBreak),
     shebang: pattern_reader.starting_with(/#!/).until_after(acorn.lineBreak),
-    xml: pattern_reader.matching(/[\s\S]*?<(\/?)([-a-zA-Z:0-9_.]+|{[\s\S]+?}|!\[CDATA\[[\s\S]*?\]\])(\s+{[\s\S]+?}|\s+[-a-zA-Z:0-9_.]+|\s+[-a-zA-Z:0-9_.]+\s*=\s*('[^']*'|"[^"]*"|{[\s\S]+?}))*\s*(\/?)\s*>/),
+    xml: pattern_reader.matching(/[\s\S]*?<(\/?)([-a-zA-Z:0-9_.]+|{[\s\S]+?}|!\[CDATA\[[\s\S]*?\]\]|)(\s+{[\s\S]+?}|\s+[-a-zA-Z:0-9_.]+|\s+[-a-zA-Z:0-9_.]+\s*=\s*('[^']*'|"[^"]*"|{[\s\S]+?}))*\s*(\/?)\s*>/),
     single_quote: templatable.until(/['\\\n\r\u2028\u2029]/),
     double_quote: templatable.until(/["\\\n\r\u2028\u2029]/),
     template_text: templatable.until(/[`\\$]/),
--- a/test/data/javascript/tests.js
+++ b/test/data/javascript/tests.js
@@ -2496,10 +2496,10 @@ exports.test_data = {
             'class Columns extends React.Component {',
             '    render() {',
             '        return (',
-            '            <a>',
+            '            <>',
             '              <td>Hello</td>',
             '              <td>World</td>',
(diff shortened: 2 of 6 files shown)


test/data/javascript/tests.js
```diff
@@ -2490,6 +2490,20 @@ exports.test_data = {
...
+        },
+        {
+          unchanged: [
+            'class Columns extends React.Component {',
+            '    render() {',
+            '        return (',
+            '            <a>',
+            '              <td>Hello</td>',
+            '              <td>World</td>',
+            '            </a>',
```
bitwiseman, line 2502: This needs to be empty to actually test your change right? 
```suggestion
            '            <>',
            '              <td>Hello</td>',
            '              <td>World</td>',
            '            </>',
```

js/src/javascript/tokenizer.js
```diff
@@ -126,7 +126,7 @@ var Tokenizer = function(input_string, options) {
     html_comment_end: pattern_reader.matching(/-->/),
     include: pattern_reader.starting_with(/#include/).until_after(acorn.lineBreak),
     shebang: pattern_reader.starting_with(/#!/).until_after(acorn.lineBreak),
-    xml: pattern_reader.matching(/[\s\S]*?<(\/?)([-a-zA-Z:0-9_.]+|{[\s\S]+?}|!\[CDATA\[[\s\S]*?\]\])(\s+{[\s\S]+?}|\s+[-a-zA-Z:0-9_.]+|\s+[-a-zA-Z:0-9_.]+\s*=\s*('[^']*'|"[^"]*"|{[\s\S]+?}))*\s*(\/?)\s*>/),
+    xml: pattern_reader.matching(/[\s\S]*?<(\/?)([-a-zA-Z:0-9_.]+|{[\s\S]+?}|!\[CDATA\[[\s\S]*?\]\]|)(\s+{[\s\S]+?}|\s+[-a-zA-Z:0-9_.]+|\s+[-a-zA-Z:0-9_.]+\s*=\s*('[^']*'|"[^"]*"|{[\s\S]+?}))*\s*(\/?)\s*>/),
```
bitwiseman, line 129: I think it is clearer to say this group is optional, but that might cause the number groups to change, right? 
```
    xml: pattern_reader.matching(/[\s\S]*?<(\/?)([-a-zA-Z:0-9_.]+|{[\s\S]+?}|!\[CDATA\[[\s\S]*?\]\])?(\s+{[\s\S]+?}|\s+[-a-zA-Z:0-9_.]+|\s+[-a-zA-Z:0-9_.]+\s*=\s*('[^']*'|"[^"]*"|{[\s\S]+?}))*\s*(\/?)\s*>/),
```

bitwiseman (changes requested):
Very nice change.

JuanFML:
While fixing this issue I saw something that could be another issue:
In https://beautifier.io/ whenever you try to beautify a code with an indent of 2 spaces to an indent with 4 spaces it is unable to beautify it correctly:
**Input: (with predefined options only with Support e4x/jsx syntax selected)**
```
class Columns extends React.Component {
  render() {
    return (
      <t>        
        <td>Hello</td>
        <td>World</td>
      </t>
    );
  }
}
```
**Output**:
```
class Columns extends React.Component {
    render() {
        return (
            <t>        
        <td>Hello</td>
        <td>World</td>
      </t>
        );
    }
}
```
Although if I changed to indent to 2 spaces per indent, the output would be unchanged from the first input which is the expected output. This is because the input is tabbed with 2 spaces. Maybe the error is because of the single tags, although I haven't looked into it yet

bitwiseman:
@JuanFML 
Yes, you're seeing #665.

test/data/javascript/tests.js
JuanFML: Oh! I didn't see I hadn't changed it back, I will change this now

bitwiseman (approved):

js/src/javascript/tokenizer.js
JuanFML: Oh I see what you mean, but yes it causes other problems

JuanFML:
> @JuanFML
> Yes, you're seeing #665.

Oh yes that it is what I encounter, I see now it was already reported

bitwiseman:
@Zetrick 
Not sure have your tried this on https://beautifier.io ? Looks good to me there.

Issue #1896, opened 2021-03-12
Beautification failing for code with Bitwise Not operator(~) 


amit-cws:
# Description
Beautification failing for code with Bitwise Not operator(~) 

# Input
The code looked like this before beautification:
```
if (stack.length > 0) {
  var thisPos = stack.indexOf(this)
  ~thisPos ? stack.splice(thisPos + 1) : stack.push(this)
  ~thisPos ? keys.splice(thisPos, Infinity, key) : keys.push(key)
  if (~stack.indexOf(value)) "value" = cycleReplacer.call(this, key, value)
}
```

# Expected Output
The code should have looked like this after beautification:
```
if (stack.length > 0) {
    var thisPos = stack.indexOf(this)
    ~thisPos ? stack.splice(thisPos + 1) : stack.push(this)
    ~thisPos ? keys.splice(thisPos, Infinity, key) : keys.push(key)
    if (~stack.indexOf(value)) "value" = cycleReplacer.call(this, key, value)
}
```

# Actual Output
The  code actually looked like this after beautification:
```
if (stack.length > 0) {
    var thisPos = stack.indexOf(this) ~thisPos ? stack.splice(thisPos + 1) : stack.push(this) ~thisPos ? keys.splice(thisPos, Infinity, key) : keys.push(key)
    if (~stack.indexOf(value)) "value" = cycleReplacer.call(this, key, value)
}
```

# Steps to Reproduce

## Environment
OS: Ubuntu 20.04.2 LTS
 

## Settings
Example:
```json
{
    "indent_size": 2,
    "preserve_newlines": true
}
```

bitwiseman:
This is easy enough to add. PRs welcome.

ansh-saini:
@bitwiseman There seems to be no activity on that PR. Can I take this up?

bitwiseman:
@ansh-saini 
Yes, feel free.

ansh-saini:
Hey @bitwiseman I've been exploring the js beautifier source code since a few hours. I can't seem to find where I should make the change. I tried a few things but no luck there. I also tried doing what @Alzamer did in their PR but that does not seem to fix the problem fully.
Although I found that this is not just an issue with the bitwise operator.
Try beautifying this on [the website](https://beautifier.io/)
```
const x = 1
~x ? "Yay" : "Nay"

let x = 1
++x ? "Yay" : "Nay"

const x = 1
!x ? "Yay" : "Nay"
```

[...]

PR #1904, opened 2021-04-13
Added handling ~ operator

Alzamer:
# Description
Added handling ~ operator. Now it behaves same as prefix increment or prefix decrement operator.

Fixes Issue: 
#1896


bitwiseman (changes requested):
@Alzamer Good start.   Still needs tests and the python implementation.

PR #1844, opened 2020-10-03
Numbers should be allowed to contain underscores.

mrmonroe:
# Description
- [x] Source branch in your fork has meaningful name (not `master`)

Fixes Issue: 
#1836 Numbers should be allowed to contain underscores.

# Before Merge Checklist 
These items can be completed after PR is created.

(Check any items that are not applicable (NA) for this PR)

- [x] JavaScript implementation
- [x] Python implementation (NA if HTML beautifier)
- [x] Added Tests to data file(s)
- [NA] Added command-line option(s) (NA if
- [NA] README.md documents new feature/option(s)

PR #1915, opened 2021-04-14
fixed issue 1823

carolinekistler:
# Description
- [ ✓ ] Source branch in your fork has meaningful name (not `main`)
Fixed if statement here: https://github.com/beautify-web/js-beautify/blob/master/js/src/html/beautifier.js#L698

Fixes Issue: 
1823: HTML formatter sends comments to a new line when the comment does not include a space as the first character

# Before Merge Checklist 

- [✓ ] JavaScript implementation
- [ NA] Python implementation (NA if HTML beautifier)
- [ NA Added Tests to data file(s)
- [ NA] Added command-line option(s) (NA if
- [NA ] README.md documents new feature/option(s)

### DIFF
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -695,7 +695,7 @@ Beautifier.prototype._set_tag_position = function(printer, raw_token, parser_tok
     }
 
     // Don't add a newline before elements that should remain where they are.
-    if (parser_token.tag_name === '!--' && last_token.type === TOKEN.TAG_CLOSE &&
+    if ((parser_token.tag_name !== '!--' || parser_token.tag_name === '!--') && last_token.type === TOKEN.TAG_CLOSE &&
       last_tag_token.is_end_tag && parser_token.text.indexOf('\n') === -1) {
       //Do nothing. Leave comments on same line.
     } else {


js/src/html/beautifier.js
```diff
@@ -695,7 +695,7 @@ Beautifier.prototype._set_tag_position = function(printer, raw_token, parser_tok
     }
 
     // Don't add a newline before elements that should remain where they are.
-    if (parser_token.tag_name === '!--' && last_token.type === TOKEN.TAG_CLOSE &&
+    if ((parser_token.tag_name !== '!--' || parser_token.tag_name === '!--') && last_token.type === TOKEN.TAG_CLOSE &&
```
bitwiseman, line 698: When does the expression ` (parser_token.tag_name !== '!--' || parser_token.tag_name === '!--')` not evaluate to `true`?   What happens if you remove this expression?

bitwiseman (changes requested):
Here is where tag_name get calculated:
https://github.com/beautify-web/js-beautify/blob/bc48f5b54556bfcb3d2baefb2902ccb44958c9ed/js/src/html/beautifier.js#L607

That is what is causing the bug - it is using whitespace as the delimiter.  That is reasonable for regular elements but not for comments.   You probably want to do a special case right there for `TOKEN.COMMENT` and check for whether you are dealing with `!--` or `CDATA`. 

It is interesting that the change you did make did not break any tests.  
Speaking of which, please add tests to this PR to cover the new behavior.  

Thanks!

js/src/html/beautifier.js
carolinekistler: It has to deal with the spacing issue that it is reading when reading the comments. It is reading the tag name that does not have a space after  as the tag name is not equal to '!== 'and vice versa. This expression covers both edge cases.

carolinekistler: How would you like me to add tests?

bitwiseman: Tests: See #1844.  You only the to update the test data file.  The `generated` files are generated by the build.

bitwiseman: @carolinekistler
Please fetch and merge the latest changes from main.  I've just removed the generated files from tracking to improve clarity.

bitwiseman:
@carolinekistler 
Do you think you'll get to adding the tests for this?

PR #1925, opened 2021-04-26
python: fix argument parsing, missing css-beautify config option

DanielChabrowski:
# Description

Please do notice that the default value of `keep_quiet` is different from js-beautify as there's no option to control it and I decided not to make it - python's part of the project doesn't seem very active anyway.

Fixes Issues: 
`-n` matching incorrect option. `("--preserve**-n**ewlines")` is not a tuple but a string.
```python3
>>> '-n' in ("--preserve-newlines")
True
>>> '-n' in ("--preserve-newlines",)
False
```

missing config option appearing if there are no changes:
```
'BeautifierOptions' object has no attribute 'keep_quiet'
```

keep_quiet didn't actually quiet `beautified FILE` message

# Before Merge Checklist 

### DIFF
--- a/python/cssbeautifier/css/options.py
+++ b/python/cssbeautifier/css/options.py
@@ -57,3 +57,4 @@ class BeautifierOptions(BaseOptions):
             self._get_boolean("space_around_combinator")
             or space_around_selector_separator
         )
+        self.keep_quiet = True
--- a/python/cssbeautifier/css/options.py
+++ b/python/cssbeautifier/css/options.py
@@ -57,4 +57,4 @@ class BeautifierOptions(BaseOptions):
             self._get_boolean("space_around_combinator")
             or space_around_selector_separator
         )
-        self.keep_quiet = True
+        self.keep_quiet = False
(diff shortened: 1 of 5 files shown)


python/cssbeautifier/css/options.py
```diff
@@ -57,3 +57,4 @@ def __init__(self, options=None):
             self._get_boolean("space_around_combinator")
             or space_around_selector_separator
         )
+        self.keep_quiet = True
```
bitwiseman, line 60: Shouldn't this default to `False`?

bitwiseman (changes requested):
I guess there are no tests for these options? If there were, would they be broken? 😢

DanielChabrowski:
> I guess there are no tests for these options? If there were, would they be broken? cry

I ran `make pytest` and nothing broke so I guess there aren't any for python implementation.

bitwiseman (approved):
I added a test that would fail before this change.  It's not perfect but it proves this issue is fixed and tries to limit it's return.

Issue #1610, opened 2019-01-03
'$' issues when use javascriptobfuscator_unpacker


lovelyelfpop:
# Description
Orignal code:
	var a = {
	    a: ',
	    b: 2,
	    currencySymbol: '$',
	    c: 3
	};

# Input
code after obfuscated
	var _0xbe98 = ["\x63\x75\x72\x72\x65\x6e\x63\x79\x53\x79\x6d\x62\x6f\x6c", "\x24"];
	var a = {};
	a['\x61'] = '\x31';
	a['\x62'] = 0x2;
	a[_0xbe98[0]] = _0xbe98[1];
	a['\x63'] = 0x3;

# Expected Output
	var a = {};
	a['\x61'] = '\x31';
	a['\x62'] = 0x2;
	a['currencySymbol'] = '$';
	a['\x63'] = 0x3;

# Actual Output
	var a = {};
	a['\x61'] = '\x31';
	a['\x62'] = 0x2;
	a['currencySymbol'] = '; //missing $'
	a['\x63'] = 0x3;;
	a['\x63'] = 0x3; // where is this line come from?

# Steps to Reproduce
`JavascriptObfuscator.unpack(code)`

immaithil:
Can I take up this issue

Dhananjay-12:
Hi, @bitwiseman  is this issue still open?

bitwiseman:
Yes, it is still open.

N-45div:
is this issue still open ?

PR #1928, opened 2021-04-28
Fixed issue #1610

Alzamer:
# Description

Fixes Issue: 
#1610


bitwiseman (changes requested):
Please add a test for the fix to ensure the code does not regress in the future.

Issue #1929, opened 2021-05-01
unlimited indent when switch { case: } without break;


bill1812:
# Description
as title.  It's used on https://beautifier.io/

# Input
The code looked like this before beautification:
```
I must insert a 'fake' break before process
break; /// fake, Delete after js-beautifier
```

# Expected Output
The code should have looked like this after beautification:
```
// beautifier work fine after added 'fake' break!
// result:
    for (var att in params) {
      if (params[has](att)) {
        if (!R._availableAttrs[has](att)) {
          continue;
        }
        var value = params[att];
        attrs[att] = value;

        switch (att) {
        case "blur":
          o.blur(value);
          break;

        case "href":
        case "title":
        case "target":
          var pn = node.parentNode;

          if (pn.tagName.toLowerCase() != "a") {
            var hl = $("a");
            pn.insertBefore(hl, node);
            hl.appendChild(node);
            pn = hl;
          }

          if (att == "target") {
            pn.setAttributeNS(xlink, "show", value == "blank" ? "new" : value);
          } else {
            pn.setAttributeNS(xlink, att, value);
          }
          break;

        case "cursor":
          node.style.cursor = value;
          break;

        case "transform":
          o.transform(value);
          break;

        case "arrow-start":
          addArrow(o, value);
          break;

        case "arrow-end":
          addArrow(o, value, 1);
          break;

        case "clip-rect":
          var rect = Str(value).split(separator);

          if (rect.length == 4) {
            o.clip && o.clip.parentNode.parentNode.removeChild(o.clip.parentNode);
            var el = $("clipPath");
            var rc = $("rect");
            el.id = R.createUUID();

            $(rc, { x: rect[0], y: rect[1], width: rect[2], height: rect[3] });

            el.appendChild(rc);
            o.paper.defs.appendChild(el);
            $(node, { "clip-path": "url(#" + el.id + ")" });
            o.clip = rc;
          }

          if (!value) {
            var path = node.getAttribute("clip-path");

            if (path) {
              var clip = R._g.doc.getElementById(path.replace(/(^url\(#|\)$)/g, E));
              clip && clip.parentNode.removeChild(clip);
              $(node, { "clip-path": E });
              delete o.clip;
            }
          }
          break;

        case "path":
          if (o.type == "path") {
            $(node, { d: value ? attrs.path = R._pathToAbsolute(value) : "M0,0" });
            o._.dirty = 1;

            if (o._.arrows) {
[...]

PR #1943, opened 2021-06-18
issue #1683: fix switch case indentation

pr0grammm:
# Issue
Wrong indentation observed when the last line in a case is a right brace.
# Input
```
switch (x) {
    case 0:
        if (y == z) {
            a();
        } else {
            b();
        }
    case 1:
        c();
}
```
# Expected Output

(same as input)
```
switch (x) {
    case 0:
        if (y == z) {
            a();
        } else {
            b();
        }
    case 1:
        c();
}
```
# Actual Output
```
switch (x) {
    case 0:
        if (y == z) {
            a();
        } else {
            b();
        }
        case 1:
            c();
}
```

# Solution 
Deindent for case/default token is skipped whenever the previous
token is a right brace.
However this logic doesn't always work as highlighted in the issue.
Added code to skip deindent only if the closing right brace corresponds to
opening left brace of case action block.

# Description
- [x] Source branch in your fork has meaningful name (not `main`)

Fixes Issue: https://github.com/beautify-web/js-beautify/issues/1683

# Before Merge Checklist 
These items can be completed after PR is created.

(Check any items that are not applicable (NA) for this PR)

- [x] JavaScript implementation
- [x] Python implementation (NA if HTML beautifier)
- [x] Added Tests to data file(s)
- [NA] Added command-line option(s) (NA if
- [NA] README.md documents new feature/option(s)

PR #1948, opened 2021-06-24
Fix issue #1929

SM-90:
# Description
- Fix for issue#1929
The issue is caused by assuming that a '}' after a case statement always means the end of case block. Modified the code to accommodate the situation where '}' marks the end of another block within the case block, like an if block. Added a variable to record the position of the last case statement, and deindent accordingly for successive case statements.

**sample input** 
```
case 1:
        if (x == 1) {
        }
case 2:
        func();
```

Fixes Issue: #1929.

- [NA ] Added command-line option(s) (NA if
- [ NA] README.md documents new feature/option(s)


bitwiseman:
Is this needed now? or does #1943 also address this?

SM-90:
> Is this needed now? or does #1943 also address this?
I wasn't aware of #1943 when I created the pull request. #1943 does address this, but I think the logic doesn't handle the case with multiple indents within a case block. Ex - 

case 0:
    if(x == 1)
        if(y == 2)
            if(z == 3)
                func();
case 1:
    func();

bitwiseman:
@SM-90 
This code looks fine on https://beautifier.io : 

```
switch (a) {
    case 0:
        if (x == 1)
            if (y == 2)
                if (z == 3)
                    func();
    case 1:
        func();
}
```

pr0grammm:
> @SM-90
> This code looks fine on https://beautifier.io not sure if #1943 changes this :
> 
> ```
> switch (a) {
>     case 0:
>         if (x == 1)
>             if (y == 2)
>                 if (z == 3)
>                     func();
>     case 1:
>         func();
> }
> ```

Hi @bitwiseman , I had verified that my PR #1943 doesn't change this behaviour...thanks

PR #1967, opened 2021-11-30
Task/normalize

bitwiseman:
# Description

Fixes Issue: 

# Before Merge Checklist 

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+*.java text eol=lf
+*.py text eol=lf
+*.js text eol=lf
+*.json text eol=lf
+*.svg text eol=lf
+*.css text eol=lf
+*.mustache text eol=lf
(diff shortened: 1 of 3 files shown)


nasvillanueva:
@bitwiseman It might be late and I'm sorry for necroposting, but may I ask what's with the addition of npm in the dependencies?

This addition broke one of our project's build (as for why after 4 months, I'm not sure, I'm using ^1.8.7 so it should've gotten it then). What happened was, we include `node_modules/.bin` to our path like so `PATH=/path/to/node_modules/.bin:$PATH` for our CI. Since node_modules/.bin preceeds our global npm path, it's opting to use the npm inside node_modules. We found it and got around it, so there's really no problem.

But I'm curious why it was added in the first place as there was no explanation on the commits or this PR.

bitwiseman:
Likely a mistake.  PR to fix is welcome.

bitwiseman:
This was merged 4 months ago. It was only released in the last few days. Fixed yesterday.

PR #1968, opened 2021-12-06
GHSL-2021-113: fix ReDoS bug

kevinbackhouse:
This regex performs very badly on strings like this:

```
str = '''
return <- {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {a} {>
'''
```

PR #1973, opened 2021-12-08
GHSL-2021-113: Fix ReDoS bug (second attempt)

kevinbackhouse:
The fix that I suggested in #1968 was incorrect - I am sorry about that. Hopefully, this one is better. I have made the same change in the `python` and `js` subdirectories.

I had to make one non-trivial change to the regex to handle this test case:

https://github.com/beautify-web/js-beautify/blob/37098bafd1d22cbc5c9f1ede30268fa86e26fea2/test/data/javascript/tests.js#L2178

The problem is that it contains nested braces: `{return <tag>{item}</tag>}`. You would need to use a proper parser, rather than a regex, to handle arbitrary nesting depths, so I just updated the regex so that it can handle one level of nesting. I did that by replacing:

```
{[\s\S]+?}
```

with:

```
{([^{}]|{[^}]+?})+?}
```


bitwiseman:
@kevinbackhouse 
Thank you so much for taking a second swing at this.  

What you describe is what we've done with javascript string matching including template strings:

https://github.com/beautify-web/js-beautify/blob/37098bafd1d22cbc5c9f1ede30268fa86e26fea2/js/src/javascript/tokenizer.js#L510-L561

And yes it is considerably more complex.  

What you've done is enough to get all the tests to pass - Yay! Still, I'm concerned that it will end up being a breaking change for a significant number of users. Could you open a new issue for the multiple nested scenario you mentioned?  Then we can merge this acknowledging the limitation of the fix and that we have some idea of how to address it.

kevinbackhouse:
@bitwiseman: I created #1974. Is that what you have in mind?

bitwiseman (approved):

bitwiseman:
Perfect, thanks!

Tzahile:
Thanks for the effort, @kevinbackhouse and @bitwiseman :)
When will you publish a new version with the fix?

bitwiseman:
It is out now.

PR #1977, opened 2021-12-26
#midterm_issue_1400

ArashST79:
I changed the last output(sweet_code) for adding whitespace in correct positions


bitwiseman:
This fix is not mergable. No tests and hacking regex on the end of processing is unmaintainable.

Issue #1991, opened 2022-03-04
[Bug] Logical assignments in JS are incorrectly beautified 


mhnaeem:
# Description
The tool incorrectly beautifies logical AND, OR and NULL assignments.

# Links:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND_assignment
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment

# Input
The code looked like this before beautification:
```
x&&=y
x||=y
x??=y
```

# Expected Output
The code should have looked like this after beautification:
```
x &&= y
x ||= y
x ??= y
```

# Actual Output
The  code actually looked like this after beautification:
```
x && = y
x || = y
x ?? = y
```

# Steps to Reproduce

1. Opened https://beautifier.io/
2. Entered the input text
3. Clicked "Beautify Code" button

I am new to the repo but would be happy to work on this issue

mhnaeem:
@bitwiseman What is the process for closing issues in the project? I noticed that we released some issues but they weren't closed like this one. In addition, what is the update process for the website (https://beautifier.io/) I noticed that this issue still doesn't work on the site.

bitwiseman:
@mhnaeem 
Thanks for calling attention to this.  It looks like github pages publishing action is failing in beautifier/beautifier.io project: https://github.com/beautifier/beautifier.io/actions/runs/2059622873

I'll take a look at it some time this week.

PR #1992, opened 2022-03-05
Added support for logical AND, NULL, and OR assignments

mhnaeem:
# Description
Added support for logical AND, NULL, and OR assignments

Fixes Issue: 
#1991

# Before Merge Checklist 
- [X] JavaScript implementation
- [X] Python implementation (NA if HTML beautifier)
- [X] Added Tests to data file(s)
- [NA] Added command-line option(s)


mhnaeem:
These tests seem to be failing due to the latest release of [psf/black](https://github.com/psf/black) package, [v22.1.0](https://github.com/psf/black/blob/main/CHANGES.md#2210) breaks the configuration. Perhaps we can specify an older version of the package for install until we can address that issue.

`$(SCRIPT_DIR)/python-dev3 pip install black==21.12b0` in `Makefile`

bitwiseman (approved):

Issue #1995, opened 2022-03-06
Specify an old version of psf/black package because it removed support for python 2


mhnaeem:
# Description
Python tests seem to be failing due to the latest release of [psf/black](https://github.com/psf/black) package, [v22.1.0](https://github.com/psf/black/blob/main/CHANGES.md#2210) which breaks the configuration because they stopped supporting python version 2.

# Fixes
Fix an old version of the package for download until the tooling can be reconfigured

PR #1996, opened 2022-03-06
bug: fix psf/black package version

mhnaeem:
# Description
The latest release of black has introduced breaking changes which affect the ci pipeline and python test. Specify a version for download until that can be fixed.

Fixes Issue: 
#1995

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ $(BUILD_DIR)/virtualenv: | $(BUILD_DIR)
 	virtualenv --version || pip install virtualenv
 	virtualenv build/python-dev
 	virtualenv build/python-rel
-	$(SCRIPT_DIR)/python-dev3 pip install black
+	$(SCRIPT_DIR)/python-dev3 pip install black==21.12b0
 	@touch $(BUILD_DIR)/virtualenv
 
 


bitwiseman:
Instead of locking the version, perhaps you could run the new version and apply the changes?

bitwiseman (changes requested):
If we can avoid locking to an old version it would be better.

mhnaeem:
> Instead of locking the version, perhaps you could run the new version and apply the changes?

@bitwiseman I am not 100% familiar with the usage of that tool but according to their release notes for the [v22.1.0](https://github.com/psf/black/blob/main/CHANGES.md#2210) they removed entire support for Python 2. So the fix here would be to use an alternative package or stop running Python2 by removing `'py27'` from the target versions in `python/pyproject.toml`.

See link for more info: https://github.com/psf/black/pull/2740

bitwiseman:
It is probably time for this tool to remove python 2 support. Would you be willing to make that change?

mhnaeem:
> Would you be willing to make that change?

@bitwiseman I am not entirely sure what is required to make that possible. I will be happy to contribute some time if there is a well defined list of issues or things that need done.

bitwiseman (approved):

Issue #1846, opened 2020-10-11
in keyword in class method causes indentation problem


laggingreflex:
# Description

Class with a function named "in" causes indentation issue:

# Input
The code looked like this before beautification:
```
class {
  get a() {

  }

  in () {

  }

  b () {

  }
}
```

# Expected Output
The code should have looked like this after beautification:
```
(should be unchanged)
```

# Actual Output
The  code actually looked like this after beautification:
```
class {
    get a() {

        }

        in () {

        }

    b() {

    }
}
```

# Steps to Reproduce

Tried on https://beautifier.io/

kcamsanc:
@bitwiseman Hi! I'm new to open source contributions and am considering addressing this issue. Before I take it though, I was wondering, is it still a requirement to ensure feature parity as described in [Contributions.md](https://github.com/beautify-web/js-beautify/blob/main/CONTRIBUTING.md)?

If so, what's the best way to reproduce this error in the Python implementation after forking and cloning the repo locally? Thanks.

kcamsanc:
> If so, what's the best way to reproduce this error in the Python implementation after forking and cloning the repo locally? Thanks.

I believe I figured out the answer to my question above.

I am now working on this change. I will keep you all updated on my progress!

kcamsanc:
Hi all, I just opened up PR #2004. Let me know what you think. There is one thing that I wanted to bring up though (also under "CURRENT DISCUSSION" in the PR):

- I'm under the impression that when I add unit tests to /test/data/*/test.js, corresponding tests are automatically generated for the Python implementation. Is this correct?
- If so, I found that my unit tests did not fail for the Python implementation before I made any changes to that implementation. However, I created a .js file myself, ran ./tools/python-dev js-beautify on my file, and found that the issue was actually reproducible in the Python implementation.
- How can I add unit tests for the Python implementation to incur a test failure before making changes to the Python implementation?

bitwiseman:
> Hi all, I just opened up PR #2004. Let me know what you think. There is one thing that I wanted to bring up though (also under "CURRENT DISCUSSION" in the PR):
> 
> * I'm under the impression that when I add unit tests to `/test/data/*/test.js`, corresponding tests are automatically generated for the Python implementation. Is this correct?

Yes.

[...]

PR #2004, opened 2022-03-29
[ISSUE 1846] Resolve indent issue with class method named in

kcamsanc:
# Description

Fixes Issue: 
#1846 Originally, if a class method was named `in`, it would get reformatted one level too far. This was caused by a previous hack for the `in` and `of` keywords, which would be tokenized as `OPERATOR`s. 

The fix was to tokenize `in` and `of` as `OPERATOR`s only if the previous token was a `WORD` or `STRING`. This was how I differentiated between `in`/`of` as a keyword or as a class method name. Note: these changes resolve the issue for the `of` keyword as well.

I added unit tests to test `in`/`of` in cases in which they are keywords and when they are class method names. 

CURRENT DISCUSSION:
- I'm under the impression that when I add unit tests to `/test/data/*/test.js`, corresponding tests are automatically generated for the Python implementation. Is this correct?
- If so, I found that my unit tests did _not_ fail for the Python implementation before I made any changes to that implementation. However, I created a `.js` file myself, ran `./tools/python-dev js-beautify` on my file, and found that the issue was actually reproducible in the Python implementation.
- How can I add unit tests for the Python implementation to incur a test failure before making changes to the Python implementation?

# Before Merge Checklist 


python/jsbeautifier/javascript/tokenizer.py
```diff
@@ -22,6 +22,8 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from ast import Str
+from lib2to3.pgen2.token import STRING
```
bitwiseman, line 26: ? Tell me about these new imports.

kcamsanc:
Hi, just checking back in, is there any more work to be done with this issue on my end?

Issue #1838, opened 2020-09-20
Overindentation when using "class" as a key in an object


laggingreflex:
Expected: 

```js
{
    class: {
        a: 1,
        b: 2,
        c: 3,
    }
}
```

Actual:

```js
{
    class: {
        a: 1,
            b: 2,
            c: 3,
    }
}
```

bitwiseman:
Another keyword needing a special case.

mhnaeem:
Fix for this issue released with #2013

PR #2013, opened 2022-04-03
'class' as a object property name

mhnaeem:
Checking to make sure that the token with text 'class' or 'interface' has whitespace after it to be considered a block statement, otherwise treat as object literal

# Description
- [X] Source branch in your fork has meaningful name (not `main`)

Fixes Issue: 
#1838

### DIFF
--- a/js/src/javascript/beautifier.js
+++ b/js/src/javascript/beautifier.js
@@ -691,10 +691,10 @@ Beautifier.prototype.handle_start_block = function(current_token) {
     )) {
     // We don't support TypeScript,but we didn't break it for a very long time.
     // We'll try to keep not breaking it.
-    if (!in_array(this._last_last_text, ['class', 'interface'])) {
-      this.set_mode(MODE.ObjectLiteral);
-    } else {
+    if (in_array(this._last_last_text, ['class', 'interface']) && !!second_token.whitespace_before) {
       this.set_mode(MODE.BlockStatement);
+    } else {
+      this.set_mode(MODE.ObjectLiteral);
     }
   } else if (this._flags.last_token.type === TOKEN.OPERATOR && this._flags.last_token.text === '=>') {
     // arrow function: (param1, paramN) => { statements }
--- a/js/src/javascript/beautifier.js
+++ b/js/src/javascript/beautifier.js
@@ -691,7 +691,7 @@ Beautifier.prototype.handle_start_block = function(current_token) {
     )) {
     // We don't support TypeScript,but we didn't break it for a very long time.
     // We'll try to keep not breaking it.
-    if (in_array(this._last_last_text, ['class', 'interface']) && !!second_token.whitespace_before) {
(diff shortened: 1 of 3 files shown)


test/data/javascript/tests.js
```diff
@@ -230,7 +230,18 @@ exports.test_data = {
...
+      tests: [{
+          comment: "#1838 - handle class word as an object property",
+          unchanged: [
+            '{\n',
+            '    class: {\n',
+            '        a: 1,\n',
+            '        b: 2,\n',
+            '        c: 3,\n',
+            '    }\n',
+            '}'
```
bitwiseman, line 242: Please move your test input to the regression test area down around line 3500 or so.  This doesn't relate to "Object literal shorthand functions".

js/src/javascript/beautifier.js
```diff
@@ -691,10 +691,10 @@ Beautifier.prototype.handle_start_block = function(current_token) {
     )) {
     // We don't support TypeScript,but we didn't break it for a very long time.
     // We'll try to keep not breaking it.
-    if (!in_array(this._last_last_text, ['class', 'interface'])) {
-      this.set_mode(MODE.ObjectLiteral);
-    } else {
+    if (in_array(this._last_last_text, ['class', 'interface']) && !!second_token.whitespace_before) {
```
bitwiseman, line 694: I don't think whitespace is a good indicator for what you're trying to do here. 

The following two inputs should produce the same output, but I don't think they do with this change:
* `var a={class:{a:1}}`
* `var a = { class : { a : 1 } }`

bitwiseman (changes requested):
Thanks for submitting this PR!
It is a good quality PR, but still needs some changes.

js/src/javascript/beautifier.js
mhnaeem: @bitwiseman Both the cases you have stated resulted in the same output using whitespace however I took your comments into consideration and changed the condition to look for the text for the `second_token`.

test/data/javascript/tests.js
mhnaeem: I wasn't sure about this, thanks for letting me know. I have moved them under regression tests and also added more tests.

test/data/javascript/tests.js
```diff
@@ -3304,6 +3293,42 @@ exports.test_data = {
...
+        },
+        {
+          comment: "#1838 - handle class word as an object property but with space after colon",
+          input: [
+            '{',
+            '    class : { a: 1,',
+            'b: 2,c : 3',
+            '    }',
+            '}'
+          ],
```
bitwiseman, line 3321: This is good.  I think you need to test at least two more inputs:
* `{class:{a:1,b:2,c:3,}}`
* `{x:{a:1,class:2,c:3,}}`

bitwiseman (changes requested):
A couple more tests and we can merge this.

bitwiseman (approved):

bitwiseman:
Great work! Thanks!

PR #2017, opened 2022-04-07
[LESS] Fixing issues with spacing when an object literal lives inside a mixin

### LINKED ISSUE #722: LESS mixins gets formatted strangely
Given code like this:

```
.mq-medium( {
    position: absolute;
    background: red;
});
```

It ends up like this by the CSS Beautifier formatter:

```
.mq-medium( {
    position: absolute;
    background: red;
}
);
```

It still works, but it just looks a bit strange that } and ); ends up on their own rows.

```
```

### LINKED ISSUE #2016: LESS mixin unexpected code formatting
# Description
This is moved over from https://github.com/microsoft/vscode/issues/146693

Using VSCode 1.66.0 and the new built-in LESS code formatting feature `less.format.enable` implemented by the [JS Beautify library](https://github.com/beautify-web/js-beautify).

When using a LESS mixin that accepts CSS code rules as a parameter (this is used for media query mixins), the formatting when saved is a little unexpected. The formatting is consistent, but it is unusal.

# Input & Expected Output

```less
.example(@rules) {
    @rules();
}

.test {
    .example({
        color:red;
    });
}
```

# Actual Output

```less
.example(@rules) {
    @rules();
}

.test {
    .example( {
            color:red;
        }

    );
}
```

# More Information

Similar issue reported a while back [js-beautify issue 722](https://github.com/beautify-web/js-beautify/issues/722)

I've noticed that if you add the following to VSCode's `settings.json` file, the results are slightly better:

```
"less.format.newlineBetweenRules": false
```

This gives you formatting like:
```less
.example(@rules) {
    @rules();
}

.test {
    .example( {
            color:red;
        }
    );
}
```

I was asked to move the issue over here, so hope it's in the right place now.

## Environment
OS: macOS 12.3.1 VSCode 1.66.0 which bundles js-beautify

I have been trying to track down the version of js-beautify, but I cannot seem to find it within VSCode, so if anyone can advise how I can report this that would be great.

## Settings
Defined by VSCode

mhnaeem:
# Description
Fix issues with spacing in LESS code when using object literals inside mixins. The changes in this PR could be considered breaking since but I believe they are correct.

Fixes Issue: 
- https://github.com/beautify-web/js-beautify/issues/722
- https://github.com/beautify-web/js-beautify/issues/2016
- https://github.com/microsoft/vscode/issues/146693

Fixes #722
Fixes #2016

### DIFF
--- a/test/data/css/tests.js
+++ b/test/data/css/tests.js
@@ -1801,7 +1801,7 @@ exports.test_data = {
           '        a: b;',
           '    }',
           '});'
-        ]    
+        ]
       }, {
         comment: 'Ensure simple closing parens do not break behavior',
         unchanged: [
--- a/test/data/css/tests.js
+++ b/test/data/css/tests.js
@@ -1793,6 +1793,16 @@ exports.test_data = {
           '    });',
           '}'
         ]
+    }, {
+        unchanged: [
+          '@selectors: blue, green, red;',
+          '',
+          'each(@selectors, {',
+          '    .sel-@{value} {',
+          '          a: b;',
+          '    }',
+          '});'
+        ]    
       }, {
         comment: 'Ensure simple closing parens do not break behavior',
         unchanged: [
--- a/test/data/css/tests.js
+++ b/test/data/css/tests.js
@@ -1360,9 +1360,8 @@ exports.test_data = {
             '}',
             '.set {',
             '    each(@set, {',
-            '            @{key}-@{index}: @value;',
-            '        }',
-            '    );',
+            '        @{key}-@{index}: @value;',
+            '    });',
             '}'
           ]
         },
@@ -1733,6 +1732,20 @@ exports.test_data = {
           '}'
         ]
       }, {
+        comment: 'mixins call with object notation, and brace_style="expand"',
+        input: [
+          '.example({',
+          '    color:red;',
+          '});'
+        ],
+        output: [
+          '.example(',
+          '    {',
+          '        color:red;',
+          '    }',
+          ');'
+        ]
+      }, {
         comment: 'integration test of newline_between_rules, imports, and brace_style="expand"',
         input: '.a{} @import "custom.css";.rule{}',
         output: [
@@ -1756,6 +1769,28 @@ exports.test_data = {
           '}',
           'strong {',
           '    &:extend(a:hover);',
+          '}'
+        ]
+      }, {
+        unchanged: [
+          '.test {',
+          '    .example({',
+          '        color:red;',
+          '    });',
+          '}'
+        ]
+      }, {
+        unchanged: [
+          '.example2({',
+          '    display:none;',
+          '});'
+        ]
+      }, {
+        unchanged: [
+          '.aa {',
+          '    .mq-medium(a, {',
+          '        background: red;',
+          '    });',
           '}'
         ]
       }, {
--- a/test/data/css/tests.js
+++ b/test/data/css/tests.js
@@ -1796,7 +1796,6 @@ exports.test_data = {
       }, {
         unchanged: [
           '@selectors: blue, green, red;',
-          '',
           'each(@selectors, {',
           '    .sel-@{value} {',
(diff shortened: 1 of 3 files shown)


test/data/css/tests.js
```diff
@@ -1360,9 +1360,8 @@ exports.test_data = {
             '}',
             '.set {',
             '    each(@set, {',
-            '            @{key}-@{index}: @value;',
-            '        }',
-            '    );',
+            '        @{key}-@{index}: @value;',
+            '    });',
```
mhnaeem, line 1364: Had to change this test because I believe it is incorrect. It adds extra indents to properties which is unexpected, this test also conflicting with my fixes. I have added example of what it would look like before my changes and after.

Before PR:
```less
@set: {
    one: blue;
    two: green;
    three: red;
}
.set {
    each(@set, {
            @{key}-@{index}: @value;
        }
    );
}
```

After PR:
```less
@set: {
    one: blue;
    two: green;
    three: red;
}
.set {
    each(@set, {
        @{key}-@{index}: @value;
    });
}
```

mhnaeem: Please refer to the example on the LESS's website which also uses the same formatting as after my changes
https://lesscss.org/functions/#list-functions-each

test/data/css/tests.js
```diff
@@ -1758,6 +1771,28 @@ exports.test_data = {
...
+          '});'
+        ]
+      }, {
+        unchanged: [
+          '.aa {',
+          '    .mq-medium(a, {',
+          '        background: red;',
+          '    });',
+          '}'
+        ]
```
bitwiseman, line 1795:
```suggestion
        unchanged: [
          '.aa {',
          '    .mq-medium(a, {',
          '        background: red;',
          '    });',
          '}'
        ]
    }, {
        unchanged: [
          '@selectors: blue, green, red;',
          '',
          'each(@selectors, {',
          '    .sel-@{value} {',
          '          a: b;',
          '    }',
          '});'
        ]    
```

test/data/css/tests.js
```diff
@@ -1758,6 +1771,38 @@ exports.test_data = {
...
+    }, {
```
bitwiseman, line 1796:
```suggestion
      }, {
```

test/data/css/tests.js
```diff
@@ -1758,6 +1771,38 @@ exports.test_data = {
...
+          '}'
+        ]
+      }, {
+        unchanged: [
+          '@selectors: blue, green, red;',
+          '',
+          'each(@selectors, {',
+          '    .sel-@{value} {',
+          '          a: b;',
+          '    }',
```
bitwiseman, line 1803:
```suggestion
          '    .sel-@{value} {',
          '        a: b;',
          '    }',
```

test/data/css/tests.js
bitwiseman: You are correct.  Sometimes the tests document existing wrong behavior so that we'll know when something is fixed.

test/data/css/tests.js
```diff
@@ -1758,6 +1771,38 @@ exports.test_data = {
...
+          '.aa {',
+          '    .mq-medium(a, {',
+          '        background: red;',
+          '    });',
+          '}'
+        ]
+      }, {
+        unchanged: [
+          '@selectors: blue, green, red;',
+          '',
```
bitwiseman, line 1799:
```suggestion
```

test/data/css/tests.js
```diff
@@ -1758,6 +1771,37 @@ exports.test_data = {
...
+        ]
+      }, {
+        unchanged: [
+          '@selectors: blue, green, red;',
+          'each(@selectors, {',
+          '    .sel-@{value} {',
+          '        a: b;',
+          '    }',
+          '});'
+        ]
```
bitwiseman, line 1804:
```suggestion
        ]
```

bitwiseman (approved):
Nicely done!

Issue #1095, opened 2017-01-01
improper line concatenation between 'return' and a prefix expression


kotarondo:
# Description

js-beautify erroneously concatenates two lines in javascript:
line1 which ends with return statement with no expression nor semicolon
and
line2 which starts with prefix expression.

for example,
line1: if(x) return
line2: ++x

 Concatenating these two lines changes meaning of the program. 

# Input
The code looked like this before beautification:
```
function test(){
if(x) return
++x
}
```

# Expected Output
The code should have looked like this after beautification:
```
function test(){
    if(x) return
    ++x
}
```

# Actual Output
The  code actually looked like this after beautification:
```
function test(){
    if(x) return ++x
}
```

## Steps to Reproduce

1. Let input.js be a file containing above input program.
2. execute following:
```
js-beautify input.js > output.js
```

## Environment
Node.js: 6.8.1
js-beautify: 1.6.4

bhaveshvasnani:
@bitwiseman could I please take this issue if no one's working on this?

bitwiseman:
@bhaveshvasnani  
Feel free to work on it. I do not mean to dissuade you, but most cases of this issue are solved, but not all.  The remaining cases tend to be ones where it is hard to determine whether the operator is prefix or postfix.  

Unless you feel compelled to work on this specific issue, there are plenty of others that I would suggest you work on before this one.

PR #2028, opened 2022-04-10
Adding missing test cases for special words followed by prefix operators on new line

mhnaeem:
# Description
Adding missing test cases for special words followed by prefix operators on new line. This is related to #1095 since fixing that issue seems to break this functionality.

### DIFF
--- a/test/data/javascript/tests.js
+++ b/test/data/javascript/tests.js
@@ -3348,6 +3348,51 @@ exports.test_data = {
           ]
         },
         {
+          comment: "#1095 - Return without semicolon followed by prefix on a new line",
+          input: [
+            'function x(){',
+            'return',
+            '++a',
+            '}',
+            '',
+            'while(true) {',
+            'return',
+            '--b',
+            '}'
+          ],
+          output: [
+            'function x() {',
+            '    return ++a',
+            '}',
+            '',
+            'while (true) {',
+            '    return --b',
+            '}'
+          ]
+        },
+        {
+          comment: "#1095 - Return with semicolon followed by a prefix on a new line",
+          input: [
+            'function x(){',
+            'return; ++a',
+            '}',
+            '',
+            'while(true){return; --b',
+            '}'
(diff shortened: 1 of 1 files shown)


test/data/javascript/tests.js
```diff
@@ -3347,6 +3347,51 @@ exports.test_data = {
...
+          output: [
+            'function x() {',
+            '    return ++a',
+            '}',
+            '',
+            'while (true) {',
+            '    return --b',
+            '}'
+          ]
+        },
```
bitwiseman, line 3372: Just to be clear this is a bug.  The output is functionally different from the input: 

```suggestion
          comment: "#1095 - Return without semicolon followed by prefix on a new line - BUG output should not pull items back to same line as return",
          input: [
            'function x(){',
            'return',
            '++a',
            '}',
            '',
            'while(true) {',
            'return',
            '--b',
            '}'
          ],
          output: [
            'function x() {',
            '    return ++a',
            '}',
            '',
            'while (true) {',
            '    return --b',
            '}'
          ]
        },
```

bitwiseman (changes requested):
I assume you are intending the first of the tests to document the current incorrect behavior.

test/data/javascript/tests.js
mhnaeem: It is functionally different in JS but if we think about a pure whitespace only formatter I would expect that it would get pulled back because it would not make sense to just have a return statement followed by anything. If this is not desired behaviour then I can fix the tests and the issue linked above.

bitwiseman: Regardless of what the formatter thinks or if the code doesn't make sense, if the formatter changes the functional behavior of the code it is not longer a formatter.  The input can be unreasonable and we still need to format it accurately.  Sorry.

mhnaeem: Understood! I will fix the issue.

PR #2032, opened 2022-04-11
Experiment Only - css tokenizer

mhnaeem:
**Do Not Merge This**

Experiment - css tokenizer

Definitely not correct code, just me experimenting with possibilities, I did not commit a lot of the code so it is lost.

### DIFF
--- a/js/src/css/beautifier.js
+++ b/js/src/css/beautifier.js
@@ -73,6 +73,20 @@ function Beautifier(source_text, options) {
 
 }
 
+Beautifier.prototype._lookBack = function(testVal) {
+  var go_back_space = testVal.length;
+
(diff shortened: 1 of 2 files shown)


js/src/css/beautifier.js
```diff
@@ -68,11 +69,53 @@ function Beautifier(source_text, options) {
     "@document": true
   };
   this.NON_SEMICOLON_NEWLINE_PROPERTY = [
+    "grid-template-areas",
```
bitwiseman, line 72: You can put this in another PR and get it merged separate from this change.

mhnaeem: @bitwiseman Done #2035

PR #2033, opened 2022-04-11
Fixes #1622 - Adds space_after_named_function option for ES6 classes

### LINKED ISSUE #1622: space_after_named_function not working inside an ES6 class
# Description
space_after_named_function not working inside an ES6 class. I don't know if this is the indented behaviour or not, but it was not what I was expecting

# Input
```
class blah {
  constructor() {
    this.doStuff()
  }

  doStuff() {
    console.log("stuff")
  }
}

function test() {}
```

# Expected Output
```
class blah {
  constructor () {
    this.doStuff()
  }

  doStuff () {
    console.log("stuff")
  }
}

function test () {}
```

# Actual Output
```
class blah {
  constructor() {
    this.doStuff()
  }

  doStuff() {
    console.log("stuff")
  }
}

function test () {}
```

# Steps to Reproduce
Run on https://beautifier.io/

## Environment
https://beautifier.io/
Also, ubuntu 18.10

## Settings
Example:
```json
{
  "indent_size": "2",
  "indent_char": " ",
  "max_preserve_newlines": "5",
  "preserve_newlines": true,
  "indent_scripts": "normal",
  "brace_style": "collapse",
  "unescape_strings": false,
  "jslint_happy": false,
  "end_with_newline": false,
  "wrap_line_length": "0",
  "indent_inner_html": false,
  "comma_first": false,
  "e4x": true,
  "space_after_anon_function": true,
  "space_after_named_function": true
}
```

mhnaeem:
# Description
Adds `space_after_named_function` option for ES6 class functions

Fixes Issue: 
- Fixes #1622


mhnaeem:
@bitwiseman This PR also adds this flag `this._flags.parent.class_start_block` which I think will be very helpful in the future to detect a bunch of cool states.

js/src/javascript/beautifier.js
```diff
@@ -185,6 +185,7 @@ Beautifier.prototype.create_flags = function(flags_base, mode) {
     inline_frame: false,
     if_block: false,
     else_block: false,
+    class_start_block: false, // class A { INSIDE HERE } or class B extends C { INSIDE HERE }
```
bitwiseman, line 188: Sorry about the multiple review passes. 

Should this be called just `class_block`?

I'm thinking there should be a `class_definition` flag as well - set when the  `class` reserved word is seen. Then that continues until a `{` is seen, at which point `class_definition` is cleared and `class_block` is set. This an example of where "look back" works, but there's enough information, that it isn't needed.

Do you see what I'm saying?

bitwiseman (approved):
What you have done works and I'm willing to accept it, but take a look at my comment and consider making the suggested changes.

mhnaeem:
> Sorry about the multiple review passes.

@bitwiseman Reviews are nice, this is where I get to learn.

> Should this be called just `class_block`?

It is intentionally called `class_start_block` because the flag is only true for the `TK_START_BLOCK` -> `{` token and and not the entire block since I didn't want to assume that the closing brace will exists for the class. Example:

```js
// missing the closing bracket but still should parse correctly.
// parent for "constructor" is `{`

class A {
     constructor(){}
```

> I'm thinking there should be a class_definition flag as well - set when the class reserved word is seen

I can take a look at this, should be straight forward I think.

> Then that continues until a `{` is seen, at which point `class_definition` is cleared and `class_block` is set. 

Is there an example of a continuing flag in the code? I am not sure how to set those for multiple tokens.

> Do you see what I'm saying?

Yes, I think you suggested changes are better. I am going to attempt the requested changes.

mhnaeem:
I am looking for some help.

Example: `class A extends B {`

So how do I define a flag for multiple tokens without looking back. So for example, when assigning `class_definition` as `true` for the above example checking for `class` and `extends` token will be easy and can be assigned `true`. However, checking that `A` and `B` are also part of the `class_definition` will require me to look back and check that the previous token is of type `RESERVED` and has text `class` or `extends`.

Similar question for `class_block` which will need to be set to `true` on `{`. Without looking back it will be hard to confirm that this is still the `class_definition` going or just a random brace.

bitwiseman:
Flags are set per scope. Look at the if-block and else-block behavior. Also look at expressions - they are started by ( and continue until a matching ) is encountered.

bitwiseman:
I'm being too picky.  This is a solid, well-tested fix.  We can address internal tweaks later.

Kassy2048:
I'm in the process of making a new PR for another feature, but one test added in this PR fails:
``` js
// Issue #1622 - class/extends as a property
var a.class = {
 ...abc(),
}
b.extends({
 bb.s(),
})
```
Is it a valid JS or TS code? In particular, "class" is a reserved keyword, so "a.class" shouldn't work (same for "extends"). Also, js-beautify considers that the block that "a.class" is set to is a BlockStatement instead of an ObjectLiteral, so the trailing comma should not be possible if this is indeed a block statement.
If that is not valid, I could replace the class declaration in my PR, but I'm not sure what to use as the replacement.

PR #2034, opened 2022-04-11
[Issue #1988] Added support to recognize Handlebars block with whitespace control

### LINKED ISSUE #1988: Fails to recognize Handlebars block with whitespace control, e.g. {{~#if true ~}}
# Description

Handlebars supports [whitespace control](https://handlebarsjs.com/guide/expressions.html#whitespace-control) by putting `~` on one or both sides of any double-brace-enclosed statement. But js-beautify fails to recognize a statement with whitespace control, resulting in incorrect indentation.

# Input
```handlebars
{{#if true}}<div><div>
{{~#if true ~}}<p>true</p>{{/if}}
</div></div>{{/if}}
```

# Expected Output
```handlebars
{{#if true}}
    <div>
        <div>
            {{~#if true ~}}
                <p>true</p>
            {{/if}}
        </div>
    </div>
{{/if}}
```

# Actual Output
```handlebars
{{#if true}}
    <div>
        <div>
            {{~#if true ~}}
            <p>true</p>
{{/if}}
</div>
</div>{{/if}}
```

# Steps to Reproduce
`html-beautify -H test.hbs`

## Environment
OS: Linux

## Settings
None

zalsaedy:
# Description

Fixes Issue: 
#1988 "Fails to recognize Handlebars block with whitespace control, e.g. {{~#if true ~}}"

Added the whitespace control character inside the `tag_check_match` for the html beautifier. The control character also affected the token for partial handlebars `{{#>` to `{{~#>` which I also added a check for. Additionally, if the character is included at the start of handlebars, the test for non single handlebars tags will start at character 3 instead of character 2.

I added unit tests checking that issue 1988 was fixed, and that previous handlebar formatting fixes would not break with the inclusion of "~" at the start.

# Before Merge Checklist 

- [NA] Python implementation (NA if HTML beautifier)

### DIFF
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -607,11 +607,11 @@ var TagOpenParserToken = function(parent, raw_token) {
       tag_check_match = raw_token.text.match(/^<([^\s>]*)/);
       this.tag_check = tag_check_match ? tag_check_match[1] : '';
     } else {
-      tag_check_match = raw_token.text.match(/^{{(?:[\^]|#\*?)?([^\s}]+)/);
+      tag_check_match = raw_token.text.match(/^{{~?(?:[\^]|#\*?)?([^\s}]+)/);
       this.tag_check = tag_check_match ? tag_check_match[1] : '';
 
-      // handle "{{#> myPartial}}
-      if (raw_token.text === '{{#>' && this.tag_check === '>' && raw_token.next !== null) {
+      // handle "{{#> myPartial}}" or "{{~#> myPartial}}"
+      if ((raw_token.text === '{{#>' || raw_token.text === '{{~#>') && this.tag_check === '>' && raw_token.next !== null) {
         this.tag_check = raw_token.next.text.split(' ')[0];
       }
     }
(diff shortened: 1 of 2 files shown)


bitwiseman (approved):

bitwiseman:
@zalsaedy 
I can't see the diff for your last commit (on my phone GitHub app) but the number of changes indicates to me something amiss. I see that the checks pass but... Could you sanity check that last commit?

zalsaedy:
Hi @bitwiseman, just went through the last diff. In the old version, I closed the list of tests of tests as "}]" (on the same line), while in the new version the "]" was separated onto a new line. For some reason, this causes the diff to believe that most of the file altered was deleted and added with new text, even though the new text matches the old one.

I just went through manually and confirmed that there were no changes except the last 2 lines due to the separation of "}]", which was intentional.

bitwiseman:
@zalsaedy 

Ah, I see. Would you mind changing that back? Put them on the same line?

zalsaedy:
Heads up: last commit does list a lot of changes, but it's due to whitespace indent issues. My text editor over-indented, but running "make ci" fixed it. I did not catch this earlier, but this was the same reason with the previous commit with the large number of changes.

bitwiseman:
The total resulting changes for the PR look good again. Thanks for taking the time to clean it up.

bitwiseman (approved):

PR #2035, opened 2022-04-11
Add grid-template-areas to NON_SEMICOLON_NEWLINE_PROPERTY list

mhnaeem:
# Description
Add grid-template-areas to NON_SEMICOLON_NEWLINE_PROPERTY list. The significance of this change is visible when we start comparing entire string tokens. Right now we compare character by character and in that case `grid-template` and `grid-template-areas` will both pass the check but when we start comparing entire tokens the property name needs to be correct.


bitwiseman (changes requested):
Needs test. Then can merge.

bitwiseman (approved):

Issue #1869, opened 2020-12-29
html formatter doesn't support handlebars partial blocks (`#>`)


Goobles:
# Description

The handlebars support in the HTML beautifier is incomplete and doesn't account for partial blocks.
(as documented here: https://handlebarsjs.com/guide/partials.html#partial-blocks)

# Input
The code looked like this before beautification:
```
{{#>row}}
    {{#>column}}
        <span>content</span>
    {{/column}}
{{/row}}
```

# Expected Output
The code should have looked like this after beautification:
```
{{#>row}}
    {{#>column}}
        <span>content</span>
    {{/column}}
{{/row}}
```

# Actual Output
The  code actually looked like this after beautification:
```
{{#>row}}
    {{#>column}}
        <span>content</span>
        {{/column}}
        {{/row}}
```

# Steps to Reproduce

## Environment
OS: Windows 10 Pro
VSCode version: 1.52.0

## Settings

```json
{
  "html.format.indentHandlebars": true,
  "html.format.enable": true,
}
```

bitwiseman:
Does this reproduce on https://beautifier.io ? 
This code looks like it handles partials: https://github.com/beautify-web/js-beautify/blob/5bfad82226f98a5f6543fbde96e479e6763fd3cd/js/src/html/beautifier.js#L613-L616 .

Goobles:
Ah it does reproduce, there's a slight nuance, depending on if there's a space between the `>` and the partial name

Without the space:
```
{{#>row}}
  {{#>column}}
    <span>content</span>
    {{/column}}
    {{/row}}
```

With the space it works:
```
{{#> row}}
  {{#> column}}
    <span>content</span>
  {{/column}}
{{/row}}
```

Goobles:
That being said, both with and without spaces show the same wrong behavior in VSCode

Since I read [here](https://github.com/microsoft/vscode/issues/56346#issuecomment-413576989) that vsc uses beautify underneath I posted my issue here. Both with the vscode builtin HTML formatter and using the Beautify extension.

bitwiseman:
@Goobles  
It is likely that the version inside vscode is out of date, so it is probably missing some fixes.  

The space fixing the issue comes from here:
https://github.com/beautify-web/js-beautify/blob/5bfad82226f98a5f6543fbde96e479e6763fd3cd/js/src/html/tokenizer.js#L71

[...]

PR #2046, opened 2022-04-15
solution to partials not seperated by spaces

a-vento:
# Description

Fixes Issue: #1869 "html formatter doesn't support handlebars partial blocks"

Created a new conditional that checks for '{{#>' and '{{~#>' at the beginning of an html block. If it finds either of these, separates the partial from the tag and properly assigns the tag.  

I added unit tests checking that issue 1869 was fixed for partials not separated by spaces as well as unit tests checking that partials separated by spaces still had correct behavior 

# Before Merge Checklist 

- [ X] Python implementation (NA if HTML beautifier)
- [ X] Added command-line option(s) (NA if

### DIFF
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -608,19 +608,18 @@ var TagOpenParserToken = function(parent, raw_token) {
       this.tag_check = tag_check_match ? tag_check_match[1] : '';
     } else {
       tag_check_match = raw_token.text.match(/^{{~?(?:[\^]|#\*?)?([^\s}]+)/);
-
       this.tag_check = tag_check_match ? tag_check_match[1] : '';
 
       // handle "{{#> myPartial}}" or "{{~#> myPartial}}"
-      if ((raw_token.text === '{{#>' || raw_token.text === '{{~#>') && this.tag_check === '>' && raw_token.next !== null) {
-        this.tag_check = raw_token.next.text.split(' ')[0];
-      } else {
-        // handle "{{#>myPartial}}" or "{{~#>myPartial}}"
-        if (raw_token.text.slice(0, 4) === ('{{#>') || raw_token.text.slice(0, 5) === ('{{~#>') && this.tag_check[0] === '>') {
-          this.tag_check = raw_token.text.split('>')[1];
+      if ((raw_token.text.startsWith('{{#>') || raw_token.text.startsWith('{{~#>')) && this.tag_check[0] === '>') {
+        if(this.tag_check === '>' && raw_token.next !== null) {
+            this.tag_check = raw_token.next.text.split(' ')[0];
+        } else {
+             this.tag_check = raw_token.text.split('>')[1];
         }
       }
     }
+    
     this.tag_check = this.tag_check.toLowerCase();
 
     if (raw_token.type === TOKEN.COMMENT) {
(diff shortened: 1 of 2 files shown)


test/data/html/tests.js
```diff
@@ -3591,6 +3591,60 @@ exports.test_data = {
         '{{/if}}'
       ]
     }]
+  }, {
+    name: "Handle Partials Not Seperated by Spaces",
+    description: "Handles partials that do not have a space before the tag",
+    template: "^^^ $$$",
+    options: [
+      { name: "nospace_partials", value: "true" }
```
bitwiseman, line 3599: This is not an option, is it?

js/src/html/beautifier.js
```diff
@@ -608,11 +608,17 @@ var TagOpenParserToken = function(parent, raw_token) {
       this.tag_check = tag_check_match ? tag_check_match[1] : '';
     } else {
       tag_check_match = raw_token.text.match(/^{{~?(?:[\^]|#\*?)?([^\s}]+)/);
+
```
bitwiseman, line 611:
```suggestion
```

js/src/html/beautifier.js
```diff
@@ -608,11 +608,17 @@ var TagOpenParserToken = function(parent, raw_token) {
...
       this.tag_check = tag_check_match ? tag_check_match[1] : '';
 
       // handle "{{#> myPartial}}" or "{{~#> myPartial}}"
       if ((raw_token.text === '{{#>' || raw_token.text === '{{~#>') && this.tag_check === '>' && raw_token.next !== null) {
         this.tag_check = raw_token.next.text.split(' ')[0];
+      } else {
+        // handle "{{#>myPartial}}" or "{{~#>myPartial}}"
+        if (raw_token.text.slice(0, 4) === ('{{#>') || raw_token.text.slice(0, 5) === ('{{~#>') && this.tag_check[0] === '>') {
+          this.tag_check = raw_token.text.split('>')[1];
```
bitwiseman, line 620: There should be a way to streamline these two checks.  Maybe use `.startsWith()` instead of `===` for the first `if` and then have a conditional inside there... 

Something like this maybe? 

```
if ((raw_token.text.startsWith('{{#>') || raw_token.text.startsWith('{{~#>')) && this.tag_check[0] === '>') {
    if(this.tag_check === '>' && raw_token.next !== null) {
        this.tag_check = raw_token.next.text.split(' ')[0];
    } else {
         this.tag_check = raw_token.text.split('>')[1];
    }
}
```

I don't have it completely clear in my head, but I feel like these are largely the same scenario.

bitwiseman (changes requested):
This is a good start.  I think it needs some tuning.

a-vento:
The changes you suggested should all be fixed in the latest commit!

test/data/html/tests.js
```diff
@@ -3591,6 +3591,60 @@ exports.test_data = {
         '{{/if}}'
       ]
     }]
+  }, {
+    name: "Corrects Partial Behavior Involving Whitespace",
+    description: "Handles partials that do not have a space before the tag",
+    template: "^^^ $$$",
+    options: [
+      { name: "nospace_partials", value: "true" }
+    ],
```
bitwiseman, line 3600: What I meant with my previous comment here, was this:

```suggestion
```

bitwiseman (approved):

bitwiseman (approved):
Sorry. One more change, but otherwise looks great.

a-vento:
Oh my bad for misunderstanding your suggestion! Should be fixed now

a-vento:
reran cli

PR #2079, opened 2022-07-08
Update config object in order to prevent infinite x-scrolling

vicsantizo:
# Description
- [X] Source branch in your fork has meaningful name (not `main`)

Fixes Issue: When pasting minified code or code that has long lines to
the CodeMirror editor the options on the right side are unreachable

# Before Merge Checklist 

- [X] Python implementation (NA if HTML beautifier)
- [X] Added Tests to data file(s)
- [X] Added command-line option(s) (NA if
- [X] README.md documents new feature/option(s)

### DIFF
--- a/web/onload.js
+++ b/web/onload.js
@@ -15,7 +15,8 @@ $(function() {
   if (the.use_codemirror && typeof CodeMirror !== 'undefined') {
 
     the.editor = CodeMirror.fromTextArea(textArea, {
-      lineNumbers: true
+      lineNumbers: true,
+      lineWrapping: true,
     });
     set_editor_mode();
     the.editor.focus();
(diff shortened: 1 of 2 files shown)


web/onload.js
```diff
@@ -15,7 +15,8 @@ $(function() {
   if (the.use_codemirror && typeof CodeMirror !== 'undefined') {
 
     the.editor = CodeMirror.fromTextArea(textArea, {
-      lineNumbers: true
+      lineNumbers: true,
+      lineWrapping: true,
```
bitwiseman, line 19: We don't want to turn on codemirror line wrapping as it might be different from the wrapping the beautifier does.

PR #2081, opened 2022-07-15
Prevent editor panel from going crazy wide

aarontgrogg:
# Description
- [ ] Adding width to the Editor panel

Fixes Issue: 
Corrects problem where pasting code with no line wraps pushes the Options panel far offscreen to the right.

# Before Merge Checklist 


bitwiseman:
Fixed in another PR.

Issue #2113, opened 2022-10-27
Allow to configure the "custom elements as inline elements" behavior


cexbrayat:
# Description

Hi team! And thank you for your work on js-beautify.

I'm part of the Vue core team, and I'm one of the maintainers of the Vue Test Utils library. We use js-beautify in VTU to make the tested HTML prettier in the output. So for example, you can test something like:

```ts
expect(wrapper.html()).toBe(`<div>
  <h1>Hello</h1>
</div>`);  // the `html()` method formats the HTML with js-beautify
```

There are thousands of tests out there that use this (VTU has 1.3 million downloads per week).

js-beautify v1.14.7 introduced a new behavior in https://github.com/beautify-web/js-beautify/commit/aa82eb9ce664abadbcb454fd7a3a4d8de53dddec which is quite breaking for us, as all elements with a dash are now wrapped differently. And VTU uses elements with a `-stub` suffix when stubbing components, so a test that would usually succeed like:

```ts
expect(wrapper.html()).toBe(`<child-stub>
  <h1>Hello</h1>
</child-stub>`);
```

now fails if we update to v1.14.7 as the output of js-beautify is now `<child-stub><h1>Hello</h1></child-stub>`.

You can see that a lot of the VTU tests are failing when trying to bump to the new version of js-beautify [here](https://github.com/vuejs/test-utils/actions/runs/3330324452/jobs/5508661822).

It would be awesome if we could configure this behavior to disable it and keep the output as it was in VTU, thus avoiding the developers that use VTU having to update all their tests, or us getting stuck on v1.14.6.

A new option like `custom_element_as_inline_element` (that could be true by default, but that we could switch to false to restore the previous behavior) would be very helpful. 

Do you think that can be possible?

bitwiseman:
PR welcome.  It might make sense to reuse the existing `inline` setting and add the ability to do pattern matching.   Up to you. 

...

Wait a minute.  Before you do any work, could you please show exactly what your input, desired output, and actual output are?  From you description it sounds like this could be a bug.  

If I go to https://beautifier.io/ and put in:
```
<child-stub>
    <h1>Hello</h1>
</child-stub>
```
This input remains unchanged. I'm not sure from your description if that is correct or not...

cexbrayat:
@bitwiseman I don't think this s a bug, sorry if I wasn't clear
- input:
```html
<div><child-stub><div>hello</div></child-stub></div>
```

- desired output (actual output of v1.14.6):
```html
[...]

PR #2101, opened 2022-09-26
Avoid layout-breaking whitespace by treating all custom elements as inline elements.

lgarron:
Custom elements are any elements whose tags contain a dash: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#high-level_view

# Description
- [x] Source branch in your fork has meaningful name (not `main`)

Fixes Issue: https://github.com/beautify-web/js-beautify/issues/1989

# Before Merge Checklist 
These items can be completed after PR is created.

(Check any items that are not applicable (NA) for this PR)

- [x] JavaScript implementation
- [x] [N/A] Python implementation (NA if HTML beautifier)
- [x] Added Tests to data file(s)
- [x] [N/A] Added command-line option(s)
- [x] [N/A] README.md documents new feature/option(s)

Issue #1668, opened 2019-04-28
`include+` and `include-` properties


alexreg:
# Description

I propose the addition of `include+` and `include-` properties in the configuration options, which respectively add and remove items from the `include` list. At the moment, only the entire list can be replaced/redefined, which is quite unergonomic and does not compose well with presets, etc.

## Bonus

It would be very useful to be able to specify a function to the option that takes an HTML node object as a parameter, and returns a boolean value to specify whether the element should be beautified (or conversely, whether it should be considered inline). Returning "undefined" would fall back to the static `include`/`include+`/`include-` properties in options.

# Input

With this new feature, I expect this configuration:

```json
"inline+": [ "li" ]
```

to be equivalent to:

```json
"inline": [ "a", "abbr", "area", "audio", "b", "bdi", "bdo", "br", "button", "canvas", "cite", "code", "data", "datalist", "del", "dfn", "em", "embed", "i", "iframe", "img", "input", "ins", "kbd", "keygen", "label", "map", "mark", "math", "meter", "noscript", "object", "output", "progress", "q", "ruby", "s", "samp", "select", "small", "span", "strong", "sub", "sup", "svg", "template", "textarea", "time", "u", "var", "video", "wbr", "text", "acronym", "big", "strike", "tt", "li" ] }
```

Likewise, I expect this configuration:

```json
"inline-": [ "a" ]
```

to be equivalent to:

```json
"inline": [ "abbr", "area", "audio", "b", "bdi", "bdo", "br", "button", "canvas", "cite", "code", "data", "datalist", "del", "dfn", "em", "embed", "i", "iframe", "img", "input", "ins", "kbd", "keygen", "label", "map", "mark", "math", "meter", "noscript", "object", "output", "progress", "q", "ruby", "s", "samp", "select", "small", "span", "strong", "sub", "sup", "svg", "template", "textarea", "time", "u", "var", "video", "wbr", "text", "acronym", "big", "strike", "tt" ] }
```

bitwiseman:
Thanks @alexreg.  Due to the way CLI options are parsed, this might need to be implemented as: 
```
"inline": [ "+", "li" ]
```
And 
```
"inline": [ "-", "a" ]
```

alexreg:
@bitwiseman Sure. I mean, that's slightly less intuitive, but if it makes the implementation a lot easier, then sounds fair. Does the bonus feature (in either form) sound doable to you too?

bitwiseman:
[...]

PR #2115, opened 2022-11-11
Add option to disable custom elements inlining

cexbrayat:
# Description

Fixes Issue: 
 #2113

# Before Merge Checklist 


sagecoretech:
Hello,

This was something i needed and have code ready to commit. I added option inline_custom_tags

What is the best way to contribute these 2 lined of code?

bitwiseman:
@cexbrayat 
One of the most common problems encountered in this library is a change to the default behavior that breaks other user scenarios not currently covered by tests.  I'm sorry.  

Generally, f a change to the default behavior is _definably more correct_, which is to say we can clearly state the old behavior is wrong and the new behavior at least "more right" we will keep the new behavior.  After quite a bit of searching, it does look like #2101 which makes all custom elements inline elements is more right, but can be wrong just as often.

- [Autonomous Custom elemements are phrasing (inline) elements](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) 
- [Autonomous Custom elemements are have a transparent content model](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts) - [Transparent content model described](https://html.spec.whatwg.org/multipage/dom.html#transparent-content-models)
- Non-autonomous custom elements inherit their behavior from their parent

Add a special option to address this foible is going result in beautifier behavior be wrong in a different way just as often as it is right. 

Rather than adding a new option like you did so far, would you be willing to implement some form of #1668 instead? As part of that change making the `inline` list regex match instead of exact match?  If you do that, you could remove the special code for elements containing a dash and leave it to people to configure the inline behavior as they see fit.  

If you can't do that then we should probably revert #2101 that introduced this problem (sorry @lgarron !).  If you choose to go this route, please add tests with comments that describe your use case so at future attempts to change this will be required to address them.  

But I really hope you will consider doing #1668. It is more work, but #1668 would have broad benefits for the many users of this project.

lgarron:
Do I understand correctly that the issue "`js-beautify` changed its default behaviour" rather than "the new default behaviour is causing incorrect output"? I don't completely understand some of the linked issues, but it sounds like they can be solved by hardcoding certain lists of elements to the HTML spec.

I'm only one person, but from my perspective this is the situation:

- VS Code uses `js-beautify` for HTML parsing out of the box. It does not have configurable settings for this.
- Without #2101 in place, `js-beautify` applies un-sound transformations and **breaks the layout of HTML pages**.

I want to be able to trust my tools to have *correct* behaviour (in this case, not breaking my code), and I want to promote the same for everyone. Custom elements are a core part of modern web development, and HTML/CSS development is already subtle enough without potential breaking changes from unexpected sources. It took me a while to track down what was even happening to my code, and I have well over a decade of experience with these DOM parsing quirks.

If #2101 is reverted, I will strongly advocate for VS Code to avoid breaking code in its default formatter, whether that means running `js-beautify` with potential custom options, or switching to another formatter that preserves layout.

bitwiseman:
@lgarron 
> If #2101 is reverted, I will strongly advocate for VS Code to avoid breaking code in its default formatter, whether that means running `js-beautify` with potential custom options, or switching to another formatter that preserves layout.

Are you intending to communicate some sort of threat here?


Is anyone on this PR willing to do at least what I asked above: 

> make the `inline` list a regex match instead of exact match? If you do that, we could remove the special code for elements containing a dash and leave it to people to configure the inline behavior as they see fit.

cexbrayat:
@bitwiseman Thanks for taking a look at the PR. I personally don't have the bandwidth to dig into what you suggested, so a revert is fine by me if this PR is not enough 👍

bitwiseman (approved):

bitwiseman:
Blarg.  I don't care anymore.

cexbrayat:
@bitwiseman Thanks for taking a look and merging it!

I may have missed something, as the option is not available in the TS types. See this small repro on the TypeScript Playground: https://www.typescriptlang.org/play?ssl=6&ssc=4&pln=5&pc=4#code/JYWwDg9gTgLgBAIwKYEMCuNgDMCectQQhwDkAVgM4C0y6muJA3AFDO0bY4B0AFjCABsAFCQA8AE2AA3AHwAJJAIERRAekmySAGjgBvZnENxgAO3FITMAPoVgALyQAuOACYtBo6YGmkVgMZoFDBEVopIIBYwFM5YKAIUSMwAvgCUjEA

I can open a PR to fix this if you let me know where to look.

PR #2132, opened 2023-01-25
Added the `hyphen_attributes` HTML beautifier option to enforce all HTML attributes are hyphenated and lowercased.

mf-pherlihy:
# Description

Fixes Issue: 

Not necessarily an issue, but an improvement. Our codebase uses frameworks that rely on attributes being hyppenated and not camelCased, and having js-beautify automatically convert them would be great.

This added a new HTML only option `hyphen_attributes` which when enabled, will convert all attributes to hyphenated lowercase. By default, this is disabled and should be a noop for all current users.

# Before Merge Checklist 

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,6 +376,7 @@ HTML Beautifier Options:
   --editorconfig                     Use EditorConfig to set up the options
   --indent_scripts                   Sets indent level inside script tags ("normal", "keep", "separate")
   --unformatted_content_delimiter    Keep text content together between this string [""]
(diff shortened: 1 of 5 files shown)


bitwiseman:
@mf-pherlihy Please merge `main` into this branch to run the current CI pipeline.

bitwiseman:
At this time, the beautifier is a whitespace-only formatter.  It doesn't change text.  Please file an issue with a link to this code if you'd like to preserve it for some future date where this might be supported.

Issue #2159, opened 2023-06-21
Invalid prettification of object with unicode escape character as object key


leumasme:
# Description

For the below seen kind of Unicode Literal as Key, js-beautify breaks the code.
The resulting code creates a `SyntaxError: Invalid or unexpected token`
I also only today learned that the input is valid syntax. It's apparently equivalent to `{"𝒶": "ascr"}` (note that "𝒶" != "a")

# Input
The code looked like this before beautification:
```
{\u{1d4b6}:"ascr"}
```

# Expected Output
The code should have looked like this after beautification:
```
{
    \u{1d4b6}: "ascr"
}
```
or
```
{
  "\u{1d4b6}": "ascr"
}
```
or
```
{
  "𝒶": "ascr"
}
```

# Actual Output
The  code actually looked like this after beautification:
```
{
    \
    u {
        1 d4b6
    }: "ascr"
}
```
(This is **invalid js** which fails to execute)

# Steps to Reproduce

Simply paste input into js-beautify with default settings (web/npm)

## Environment

Tested both in browser via https://beautifier.io/
and using the npm package `js-beautify@1.14.8` on Windows 10.

## Settings
Example:
```json
{
  "indent_size": "4",
  "indent_char": " ",
  "max_preserve_newlines": "5",
  "preserve_newlines": true,
  "keep_array_indentation": false,
  "break_chained_methods": false,
  "indent_scripts": "normal",
  "brace_style": "collapse",
  "space_before_conditional": true,
  "unescape_strings": false,
  "jslint_happy": false,
  "end_with_newline": false,
  "wrap_line_length": "0",
  "indent_inner_html": false,
  "comma_first": false,
  "e4x": false,
  "indent_empty_lines": false
}
```

hilalsidhic:
Hi can someone assign this to me

bitwiseman:
@hilalsidhic Done.

Himang02:
Is the issue Open? I want to contribute.

bitwiseman:
It is open.  @hilalsidhic  doesn't seem to be working on it. Go ahead.

Chaitanyaprasad60:
Can you assign me this?

Bharrguv:
Is this still open ? If yes I am willing to contribute @bitwiseman

PR #2189, opened 2023-08-26
Handle ES6 unicode code point syntax

roja-a-m:
# Description

Fixes Issue: Invalid prettification of object with unicode escape character as object key [#2159](https://github.com/beautify-web/js-beautify/issues/2159)

# Before Merge Checklist 


bitwiseman (approved):
Thanks for submitting a PR! 

What you wrote worked within the specific case you were trying to address but didn't cover a number of other places where ES6  code points might appear. 

I updated to handle the more general case.

PR #2238, opened 2024-02-04
removed check op < is first token

sandy0096:
# Description

Fixes Issue: 

# Before Merge Checklist 

- [NA] Added Tests to data file(s)


bitwiseman:
This is present to allow there to be an html comment at the top of a javascript `<script>` element inside an html file.  Technically, your change is going in the right direction - html comments should be allowed anywhere inside a `<script>` element inside an html file.  However, it opens up a big can of potential bugs.  The tests all pass, but that doesn't mean that no existing code would be broken by this.  

At minimum, you need to provide a test case that shows this new functionality working (a test that would fail if run without this change),  Preferably, you also add some tests for inputs that are challenging, such as `a<!--b?1:0;` .  This code is unlikely, but not impossible.  

For full correctness, what we need is an "templating" called "HTML" that turns on html parsing in JavaScript when inside a `<script>` element inside an html file.  But that is ton more work and I don't expect you to do that.

bitwiseman:
@sandy0096  Do you plan to add the test?

PR #2242, opened 2024-02-15
Update index.html

NoguchiMasafumi:
I have always found your service to be very convenient. However, I have noticed that sometimes when I paste text, it can be difficult to read. Therefore, I would like to suggest that you consider implementing a spellcheck function.

# Description

Fixes Issue: 

# Before Merge Checklist 

### DIFF
--- a/index.html
+++ b/index.html
@@ -225,7 +225,7 @@
   </div>
 
   <div class="editor">
-    <textarea id="source"></textarea>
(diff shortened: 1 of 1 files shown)


bitwiseman:
I assume you mean, "turn off the spellcheck function" rather than "implement". 

For reference: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

PR #2247, opened 2024-02-16
Turn new angular templating off by default in html

### LINKED ISSUE #2246: Perf regression in latest release (1.15.0)
# Description

Overnight, the latest release (`1.15.0`) seemed to have caused a performance regression. Our "htmlPretty" function went from finishing in a few ms to taking 20s to complete. We've pinned our dependency to the last known good version that worked for us (`1.14.9`) to resolve the issue.

# Input

```
export async function htmlPretty(
  html: string,
  options?: HtmlPrettyOptions
): Promise<string> {
  const prettyOptions = options || {
    indent_size: 0,
    end_with_newline: true,
    extra_liners: [],
  };
  try {
    const output = beautify.html(html, prettyOptions);
    return output.trimStart();
  } catch (e) {
    console.error('failed to pretty html:', e);
    return html;
  }
}
```

# Expected Output

Expected to finish within X ms

# Actual Output

The function is taking about 20s to complete, and the returned HTML is not pretty-printed.

bitwiseman:
# Description
The new angular templating caused unintended side-effects when turned on by default.  This changes it by off by default.

Fixes #2246 

# Before Merge Checklist 

### DIFF
--- a/python/jsbeautifier/tests/core/test_options.py
+++ b/python/jsbeautifier/tests/core/test_options.py
@@ -102,15 +102,13 @@ class TestOptions(unittest.TestCase):
 
     def test__get_selection_list(self):
         # should raise error with empty selection
-        with self.assertRaisesRegexp(
-            ValueError, "Selection list cannot" + " be empty."
-        ):
+        with self.assertRaisesRegex(ValueError, "Selection list cannot" + " be empty."):
             Options()._get_selection_list("a", [])
         # should raise error with invalid default
-        with self.assertRaisesRegexp(ValueError, "Invalid Default Value!"):
+        with self.assertRaisesRegex(ValueError, "Invalid Default Value!"):
             Options()._get_selection_list("a", ["a", "b"], ["c"])
         # should raise error with invalid option
-        with self.assertRaisesRegexp(
+        with self.assertRaisesRegex(
             ValueError, "^Invalid Option Value:" + " The option"
         ):
             Options({"a": ["c", "d"]})._get_selection_list("a", ["a", "b"], ["a", "b"])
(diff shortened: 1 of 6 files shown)


python/jsbeautifier/tests/core/test_options.py
```diff
@@ -102,15 +102,13 @@ def test__is_valid_selection(self):
 
     def test__get_selection_list(self):
         # should raise error with empty selection
-        with self.assertRaisesRegexp(
-            ValueError, "Selection list cannot" + " be empty."
-        ):
+        with self.assertRaisesRegex(ValueError, "Selection list cannot" + " be empty."):
```
bitwiseman, line 105: These are aliases that were retained for backward compat to python 2.x apparently.  In typical python style, they just suddenly stopped working on my system.  All versions of 3 support the modern names.

liesahead:
@bitwiseman , is it possible to enable this setting currently in VSCode via `htmlFormatter` options? Because currently VSCode setting accepts a boolean and doesn't list `angular` as one of the items (https://code.visualstudio.com/docs/languages/html#_formatting).

bitwiseman:
@liesahead
HTML formatting in vscode is controlled by https://github.com/microsoft/vscode-html-languageservice . That will need to be updated to pull in the new version.  

https://github.com/microsoft/vscode-html-languageservice/blob/81cf18fc8aca3ac4ea1e5a99864731a3757260be/package.json#L24

It looks like the settings will need to be updated to allow supplying a list of templating formatters and add `angular` as an option: 

https://github.com/microsoft/vscode-html-languageservice/blob/81cf18fc8aca3ac4ea1e5a99864731a3757260be/src/htmlLanguageTypes.ts#L31-L48

https://github.com/microsoft/vscode-html-languageservice/blob/81cf18fc8aca3ac4ea1e5a99864731a3757260be/src/beautify/beautify-html.d.ts#L119

I've opened https://github.com/microsoft/vscode-html-languageservice/issues/177 requesting update.

PR #2280, opened 2024-06-17
fix(cli): templating arguments getops

joshuachp:
# Description

Fixes Issue: 

The `templating` CLI parameter requires an argument for the template to use, but this cannot be passed since the `getops` definition is missing the `=` sign.

# Before Merge Checklist 


joshuachp:
I'm not sure if this is applicable since the python version doesn't support HTML. Reopening anyway

bitwiseman:
`templating` also exists in the javascript beautifier.

bitwiseman (approved):

PR #2291, opened 2024-07-04
one-line bug in templatablepattern.py

jkmnt:
# Description
- [ -] Source branch in your fork has meaningful name (not `main`)

Fixes Issue: 

The templatablepattern.py contained the bug:
```python
    def read(self):
        if self._until_after:
            result += self._input.readUntilAfter(self._until_after) # should be self._until_pattern
```

the self._until_after is bool flag, while the `readUntilAfter` accepts the regexp.
The same code in templatablepattern.js uses the this._until_pattern correctly.

# Before Merge Checklist 

- [- ] Python implementation (NA if HTML beautifier)

### DIFF
--- a/python/jsbeautifier/core/templatablepattern.py
+++ b/python/jsbeautifier/core/templatablepattern.py
@@ -115,7 +115,7 @@ class TemplatablePattern(Pattern):
             next = self._read_template()
 
         if self._until_after:
-            result += self._input.readUntilAfter(self._until_after)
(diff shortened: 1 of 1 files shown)


bitwiseman:
The fact that this doesn't cause test failures means this code must not be exercised in Python.

bitwiseman:
Ah, I see. You  discovered this as part of porting the  HTML beautifier.

PR #2302, opened 2024-07-29
pinned setuptools version to avoid v72 since it removed a feature we're using to build the package

DanielLeone:
# Description

Fix for https://github.com/beautifier/js-beautify/issues/2301 .The current package is broken because of unpinned setuptools version

I believe we're hitting this line of the `pip` source code which is using a default `setuptools` version:
https://github.com/pypa/pip/blob/8c7df92eca1d927bb441a6b15e92966ee553aca9/src/pip/_internal/pyproject.py#L132

So this PR avoids that code path by fixing setuptools to `71.1.0` (the latest release before 72.0.0)

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,3 +1,7 @@
+[build-system]
+# pinned at v71 since v72 removed the test command https://setuptools.pypa.io/en/stable/history.html#v72-0-0
+requires = ["setuptools==71.1.0"]
+
 [tool.black]
 target-version = ['py33', 'py36', 'py37', 'py38']
 exclude = '''


cclauss (approved):
Proof that this is required in #2303

cjwatson:
#2304 is an alternative approach, which I think is better (but then I would).

MattExact:
This seems like a bit of a temporary fix.
Given this has been deprecated I think there should at the very least be a plan to remove the dependency on `setuptools.command.test` in the near future.

It looks like `tox` is recommended as a replacement:

> Users looking for a generic test entry point independent of test runner are encouraged to use tox.

But a maintainer will probably have a better idea than me of what `setuptools.command.test` is being used for here and what is a suitable replacement.

cjwatson:
While `tox` is my default, in this case `make pytest` is already documented and runs a superset of all the tests that were run by the setuptools test integration.  The latter looks vestigial to me.

cclauss:
@MattExact Please see:
* #2304

Issue #1404, opened 2018-05-31
HTML: Proposal to organize the attribute-wrapping settings


vvs:
Extracted from #1297.

## What we have

`wrap_attributes`:

* `auto` - enables wrapping of attributes.
* `force` - first attribute stays, others wrapped with "wrap_attributes_indent_size"
* `force-aligned` - the first attribute stays, others wrapped and aligned with the first one
* `force-expand-multiline` - If there are more than 1 atribute, *ALL* attributes are wrapped no matter what and indented with "wrap_attributes_indent_size". AND: the closing '>' of the opening tag is also on the new line. This mode is currently broken, produces wrongly indented files :(

There are also multiple effort to provide some additional modes that deal with wrapping of long html elements with multiple attributes: #1285 and #1262.

I think that there are multiple (orthogonal) concerns here and putting them all into single value would create more confusion and weird option names explosion if we are to handle all cases.

So, let's separate wrapping, aligning and number of attributes per line, etc.

## Proposal

`wrap_attributes_mode`:

* `none` - We just don't wrap attributes, period. (needed?)
* `auto` - Normal attribute wrapping, this should be the default, probably, as the most logical case. When the line is too big, we wrap. If all attributes can fit into a single line, they stay on the single line, no wrapping. (Default.)
* `force` - We forcefully wrap if there are more than 1 attribute present, otherwise auto.

When the wrapping is in action, the following properties are in play. If no wrapping happens, they don't play any role.

`wrapped_attributes_per_line`:

* `multiple` - Soft wrapping, *default*. Fit as many attrs per line as allowed.
* `single` - The first attribute stays on the first line, others are on separate lines.
* `single-all` - All attributes, including the first one, placed on separate lines. So, every attribute is on a separate line, and the element/tag line has no attributes left.

`wrapped_attributes_indent`:

1. auto -- Uses `wrap_attributes_indent_size`, *default*.
2. aligned -- All attribute *lines* start aligned with the first attribute. If the first attribute is on a new line ("single-all" mode of `wrapped_attributes_per_line`), then we use `wrap_attributes_indent_size` to calculate the indent.

`wrapped_attributes_end`:

* `auto` - *default*
* `multiline` - the closing `>` is on the separate line.

## Examples 

1. DEFAULT case: We soft wrap and try to fit as many attributes per line as possible, but within the line length limit.
```
wrap_attributes_mode = auto
wrapped_attributes_per_line = multiple
wrapped_attributes_indent = auto
[...]

PR #2316, opened 2024-09-28
feat: add option `closing-bracket-newline` for html beautify

zhaojjiang:
This option is added to implement prettier's `--bracket-same-line` behavior.

Input code:
```html
<div
  style="display: block;"
  width="100" height="200">
  aaaa
  <p>test content</p>
</div>
```

Output code:
```html
<div
  style="display: block;"
  width="100"
  height="200"
>
  aaaa
  <p>test content</p>
</div>;
```

# Description

# Before Merge Checklist 

### DIFF
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -376,7 +376,10 @@ Beautifier.prototype._handle_tag_close = function(printer, raw_token, last_tag_t
   } else {
     if (last_tag_token.tag_start_char === '<') {
       printer.set_space_before_token(raw_token.text[0] === '/', true); // space before />, no space before >
-      if (this._is_wrap_attributes_force_expand_multiline && last_tag_token.has_wrapped_attrs) {
+      // Add newline before tag closing bracket:
+      // 1. add newline only if 'force-expand-multipand' or 'closing-bracket-newline' is specified
+      // 2. add newline only if tag has wrapped_attrs
+      if ((this._is_wrap_attributes_force_expand_multiline || this._options.closing_bracket_newline) && last_tag_token.has_wrapped_attrs) {
         printer.print_newline(false);
       }
     }
--- a/README.md
+++ b/README.md
@@ -391,6 +391,7 @@ HTML Beautifier Options:
   --unformatted_content_delimiter    Keep text content together between this string [""]
   --indent-empty-lines               Keep indentation on empty lines
   --templating                       List of templating languages (auto,none,django,erb,handlebars,php,smarty,angular) ["auto"] auto = none in JavaScript, all in html
+  --closing-bracket-newline          Add a newline before tag closing brackets
 ```
 
 ## Directives
--- a/js/src/cli.js
+++ b/js/src/cli.js
@@ -136,7 +136,8 @@ var path = require('path'),
         "quiet": Boolean,
         "type": ["js", "css", "html"],
         "config": path,
-        "editorconfig": Boolean
+        "editorconfig": Boolean,
+        "closing_bracket_newline": Boolean
     },
     // dasherizeShorthands provides { "indent-size": ["--indent_size"] }
     // translation, allowing more convenient dashes in CLI arguments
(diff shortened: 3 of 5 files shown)


js/src/html/beautifier.js
```diff
@@ -376,7 +376,10 @@ Beautifier.prototype._handle_tag_close = function(printer, raw_token, last_tag_t
   } else {
     if (last_tag_token.tag_start_char === '<') {
       printer.set_space_before_token(raw_token.text[0] === '/', true); // space before />, no space before >
-      if (this._is_wrap_attributes_force_expand_multiline && last_tag_token.has_wrapped_attrs) {
+      // Add newline before tag closing bracket:
+      // 1. add newline only if 'force-expand-multipand' or 'closing-bracket-newline' is specified
```
bitwiseman, line 380:
```suggestion
      // 1. add newline only if 'force_expand_multiline' or 'closing-bracket-newline' is specified
```

zhaojjiang: fixed

bitwiseman:
I'm not directly opposed to adding another config option for this setting, but this project has too many options already. 

Similar to #1404 ... But for `brace-style` (which is currently not used in the html formatter except to be passed through to `<script>` contents).

zhaojjiang:
I understand your concerns a little bit.

I opened this pr because js-beautify is the basic dependency of vscode built-in html-language-feature extension which used to format html code.

There is no other way to let vscode add this option, the only way I found is open a pr in js-beautify and then wait for vscode update its dependency!

I think this project should not be so cautious aboud adding new features, as mentioned above, this project is now a built-in feature of vscode, many html format issues in vscode are linked to this project.

bitwiseman:
@zhaojjiang  
The fact that this project is bundled in vscode means it needs to be even more cautious to not introduce breaking changes.

Do you want to take responsibility for potentially breaking millions of users?

js/src/html/beautifier.js
```diff
@@ -376,7 +376,10 @@ Beautifier.prototype._handle_tag_close = function(printer, raw_token, last_tag_t
   } else {
     if (last_tag_token.tag_start_char === '<') {
       printer.set_space_before_token(raw_token.text[0] === '/', true); // space before />, no space before >
-      if (this._is_wrap_attributes_force_expand_multiline && last_tag_token.has_wrapped_attrs) {
+      // Add newline before tag closing bracket:
+      // 1. add newline only if 'force_expand_multiline' or 'closing-bracket-newline' is specified
+      // 2. add newline only if tag has wrapped_attrs
+      if ((this._is_wrap_attributes_force_expand_multiline || this._options.closing_bracket_newline) && last_tag_token.has_wrapped_attrs) {
```
bitwiseman, line 382: This makes the behavior of this new option the same as `force_expand_multiline` ...  If that is the case you don't need the new option.  I think what you want is 
```suggestion
      if ((this._is_wrap_attributes_force_expand_multiline && last_tag_token.has_wrapped_attrs)  || this._options.closing_bracket_newline) {
```

But that might not be it either...

zhaojjiang: > This makes the behavior of this new option the same as `force_expand_multiline` ... If that is the case you don't need the new option.

Yes, it is. This new option should work only with a tag which has wrap attributes, it should not work when tag has no attributes.

zhaojjiang:
hi, any plan for this option?

nowadays, I got another scene:

I use vue.volar as default vue formatter, and vue.volar format vue template code based on vscode built-in html-language-features -- which based js-beautify.
And then, I got conflict between vue eslint rule `vue/html-closing-bracket-newline` and `html-format.wrapAttributes`.

Eslint and prettier both support closing bracket newline control, And I hope js-beautify and vscode support this option too. -- I dont want to use prettier for some reasons, so I hope to solve all format rules by eslint and vscode build-in formatters.

bitwiseman (changes requested):
@zhaojjiang  
Please change the name of the option to `element-brace-style` and the possible values to `collapse` and `end-expand`.  The behavior will be same as what you have but will the possibility of additional values later.   

Then I can merge.

zhaojjiang:
for most cases and developers,

```html
<
  div
/>
```

is not an acceptable format, so we can ignore it.

In my opinion, `element-brace-style` should have only two state, so `collapse` and `expand/end-expand` is ok, the two words just opposite.

I'm not sure which additional values may be appended to this option later. If no, I prefer previous option name and boolean value, just a suggestion.

PR #2383, opened 2025-08-17
fix the issue Handlebars {{#tr}} helper is indented as if it were a <tr> tag

### LINKED ISSUE #2045: Handlebars {{#tr}} helper is indented as if it were a <tr> tag
# Description

Zulip defines a custom Handlebars helper `{{#tr}}` for translation, but when it happens to be [used inside a table cell](https://github.com/zulip/zulip/blob/0df4d8eb172bee9988d0a33c251368d2c2aa08ae/static/templates/search_operators.hbs#L15), js-beautify seemingly confuses it with a `<tr>` (table row) tag and indents it at the wrong level.

# Input
```handlebars
<table>
    <tr>
        <td>
            {{#tr}}translated{{/tr}}
        </td>
    </tr>
</table>
```

# Expected Output
```handlebars
<table>
    <tr>
        <td>
            {{#tr}}translated{{/tr}}
        </td>
    </tr>
</table>
```

# Actual Output
The code actually looked like this after beautification:
```handlebars
<table>
    <tr>
        <td>
    {{#tr}}translated{{/tr}}
    </td>
    </tr>
</table>
```

# Steps to Reproduce
`html-beautify -H test.hbs`

## Environment
OS: Linux

## Settings
None

Ankesh-Kushwaha:
Fixes #2045

# Description
 what’s happening here is that js-beautify is treating the custom Handlebars helper {{#tr}} as if it were an HTML <tr> element, so it de-indents/reindents incorrectly.

The fix is to tell js-beautify that {{#tr}} is not an HTML tag, but rather a custom tag/inline block. Luckily, js-beautify supports a custom_elements (or custom-elements) configuration option that lets you whitelist custom tags/blocks.

Fixes Issue: #2045

Solution :
Added a .jsbeautifyrc 
{
  "html": {
    "indent_inner_html": true,
    "indent_handlebars": true,
    "extra_liners": [],
    "custom_elements": ["tr"]
  }
}
# Before Merge Checklist 

### DIFF
--- a/jsbeautifyrc
+++ b/jsbeautifyrc
@@ -18,5 +18,11 @@
     "wrap_line_length": 0,
     "css": {
         "selector_separator_newline": false
-    }
+    },
+    "html": {
+    "indent_inner_html": true,
+    "indent_handlebars": true,
+    "extra_liners": [],
+    "custom_elements": ["tr"]
+  }
 }


TheRealRajdeep:
Was this PR ignored? Because I am looking to contribute to the same issue

PR #2420, opened 2026-04-02
Add "flatten" support for JS, CSS, and HTML with corresponding tests

b-balakrishna:
⚠️ Notes
This approach is intentionally minimal and matches "singular-line" key requirement.
It is not a full parser-level production minifier; whitespace flattening after beautify was added as requested.
Behavior for complicated strings/comments is “best effort” with this lightweight minify path.


bitwiseman:
This feature is not minifying, it is "one-line-ing" or flattening, as you said.

js/src/cli.js
```diff
@@ -357,6 +358,7 @@ function usage(err) {
         '  -r, --replace    Write output in-place, replacing input',
         '  -o, --outfile    Write output to file (default stdout)',
         '  --config         Path to config file',
+        '  --minify         Minify to one line output (whitespace reduction)',
```
bitwiseman, line 361: You can't use the word minify for this feature.  Sorry, but it is a complete misnomer.

PR #2442, opened 2026-05-04
Codex/issue 2386 glob security

SarthakDudhe:
Updated the project’s glob dependency from ^10.5.0 to ^11.1.0 to address the reported security vulnerability. Regenerated package-lock.json so the lockfile now resolves to glob@11.1.0 with its updated transitive dependencies. This is a dependency-only fix, so no application code changes were needed.


Jognt:
Any news on this? We tracked down a vulnerability potential in Glob@10.5.0 coming from js-beautify.
We're currently overriding it on our end, but if this is a clean fix then that would be lovely.

https://github.com/beautifier/js-beautify/pull/2427 is a PR that updates to a fully non-vulnerable version*

PR #2448, opened 2026-05-05
Codex/issue 2045 handlebars tr

SarthakDudhe:
For #2045, I fixed the HTML beautifier so Handlebars helpers like {{#tr}}...{{/tr}} are no longer mistaken for real HTML table tags during optional end-tag handling.

Description text:

Fixed a table-formatting edge case in the HTML beautifier where Handlebars helpers whose names match HTML table elements, especially {{#tr}}, were being treated like actual <tr> tags. In js/src/html/beautifier.js, _do_optional_end_element() now only applies HTML optional-closing rules when the parser token came from a real HTML tag (tag_start_char === '<'). That prevents template tokens from triggering the table-specific pop logic while preserving normal behavior for real HTML markup. I also added a regression case in test/data/html/tests.js as its own issue-specific group with indent_handlebars disabled, so the formatter keeps {{#tr}}translated{{/tr}} correctly nested inside the table cell. I verified the formatter output now stays as expected for the Handlebars helper and that regular table formatting is unchanged.

### DIFF
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -818,92 +818,93 @@ Beautifier.prototype._do_optional_end_element = function(parser_token) {
   // https://www.w3.org/TR/html5/syntax.html#optional-tags
   if (parser_token.is_empty_element || !parser_token.is_start_tag || !parser_token.parent) {
     return;
-
   }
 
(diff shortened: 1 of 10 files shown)


bitwiseman:
@SarthakDudhe  

PR #2450, opened 2026-05-05
Codex/issue 2389 django attr literal

SarthakDudhe:
Fixed #2389 by stopping the HTML tokenizer from treating Django syntax as template markup inside quoted attribute values. In js/src/html/tokenizer.js (line 62), I left the general templating scanner unchanged, but added a quoted_templatable_reader that excludes django for single_quote and double_quote attribute parsing. That means a literal string like "{#" inside attr="..." is now read as plain text instead of being mistaken for a Django comment opener that consumes the rest of the file. This keeps the formatter from collapsing the document after the attribute and preserves normal tag indentation.

I also added a regression test in test/data/html/tests.js (line 1723) for the exact upstream repro. The new case verifies that <div attr="{#"><div></div></div> now formats into a normal nested block instead of leaving the rest of the file trapped inside the attribute value. I verified the formatter output locally, and node test/generate-tests.js runs cleanly.

### DIFF
--- a/js/src/html/beautifier.js
+++ b/js/src/html/beautifier.js
@@ -818,92 +818,93 @@ Beautifier.prototype._do_optional_end_element = function(parser_token) {
   // https://www.w3.org/TR/html5/syntax.html#optional-tags
   if (parser_token.is_empty_element || !parser_token.is_start_tag || !parser_token.parent) {
     return;
-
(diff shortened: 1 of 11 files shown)


bitwiseman:
@SarthakDudhe  

PR #2453, opened 2026-05-05
Issue 2381 preserve single newlines

### LINKED ISSUE #2381: Preserve mid-value newlines surrounding parentheses
# Description
I've looked through the issues and found _similar_ posts, but nothing quite like this. I also tried looking at the code itself for a little bit, to see if I could figure out _why_ this happens, but couldn't really get my head around it.

**As far as I can tell, the formatter only allows mid-value newlines when they follow commas.** This is sensible, but in the case of nested functions specifically, I'd like to be able to insert newlines around parentheses without the formatter eating them; they indicate a boundary just like commas do, after all.

I don't THINK an implementation like this would be disruptive or horribly complex, but I know all too well how famous those last words are.

<br>

I'll also note the formatter's addition of two tabs/eight spaces, even though I don't mind it, and wouldn't mind if it kept doing that.
Maybe it's a clue as to why the newlines are consumed/where they are instructed to be.

# Input
```css
:root
{
    --size: calc(
        max(var(--foo), 
            var(--bar)
        ) + var(--padding) * 2);
}
```

# Expected Output
```css
:root
{
    --size: calc(
        max(var(--foo), 
            var(--bar)
        ) + var(--padding) * 2);
}
```
**OR**
```css
:root
{
    --size: calc(
                max(var(--foo), 
                    var(--bar)
                ) + var(--padding) * 2);
}
```

# Actual Output
The code actually looked like this after beautification:
```css
:root
{
    --size: calc(max(var(--foo),
                var(--bar)) + var(--padding) * 2);
}
```

# Steps to Reproduce

## Environment
https://beautifier.io/

_(I use VSCode on a Windows 11 machine, and found this repo through investigation along those lines. But taking inspiration from another issue I saw, I reproduced this behavior with the above online tool, for simplicity's sake.)_

## Settings
_(beautifier.io default, save for `"brace_style": "expand"`)_
```json
{
  "indent_size": "4",
  "indent_char": " ",
  "max_preserve_newlines": "5",
  "preserve_newlines": true,
  "indent_scripts": "normal",
  "brace_style": "expand",
  "unescape_strings": false,
  "jslint_happy": false,
  "end_with_newline": false,
  "wrap_line_length": "0",
  "indent_inner_html": false,
  "comma_first": false,
  "e4x": false,
  "indent_empty_lines": false
}
```

SarthakDudhe:
This one fixes #2381 by preserving single newlines that appear inside parenthesized JavaScript expressions instead of flattening them away.
What changed:
In js/src/javascript/beautifier.js, handle_whitespace_and_comments() now treats a single newline as significant when the formatter is inside an expression and preserve_newlines is enabled.
The same logic was mirrored in python/jsbeautifier/javascript/beautifier.py so the Python formatter stays aligned with the JS implementation.
I added a regression test in test/data/javascript/tests.js for:
foo(bar\nbaz); which now stays split across lines instead of collapsing to foo(bar baz);
foo(bar\n); which remains preserved as-is
The generated Python fixture picked up the same case, but that file is gitignored in this repo so you do not need to stage it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ HTML Beautifier Options:
   -M, --wrap-attributes-min-attrs    Minimum number of html tag attributes for force wrap attribute options [2]
   -i, --wrap-attributes-indent-size  Indent wrapped attributes to after N characters [indent-size] (ignored if wrap-attributes is "aligned")
   -d, --inline                       List of tags to be considered inline tags
-  --inline_custom_elements           Inline custom elements [true]
+  --inline_custom_elements           Inline custom elements [false]
   -U, --unformatted                  List of tags (defaults to inline) that should not be reformatted
   -T, --content_unformatted          List of tags (defaults to pre) whose content should not be reformatted
   -E, --extra_liners                 List of tags (defaults to [head,body,/html] that should have an extra newline before them.
--- a/README.md
+++ b/README.md
@@ -383,6 +383,7 @@ HTML Beautifier Options:
   -i, --wrap-attributes-indent-size  Indent wrapped attributes to after N characters [indent-size] (ignored if wrap-attributes is "aligned")
   -d, --inline                       List of tags to be considered inline tags
   --inline_custom_elements           Inline custom elements [false]
+  --preserve_self_closing_tags       Preserve the no-space form of self-closing tags [false]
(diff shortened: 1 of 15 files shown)


bitwiseman:
@SarthakDudhe  

PR #2469, opened 2026-06-03
Place closing brace on its own line for comment-only blocks

sarathfrancis90:
# Description

When a block or object literal contains only block comment(s) and no other tokens, the closing brace was kept on the same line as the last comment. The output was not idempotent: beautifying the output a second time moved the brace onto its own line, so `beautify(beautify(x)) !== beautify(x)`.

### Example

Input:

```js
var a = {/* c */}
```

Current output (one pass):

```js
var a = {
    /* c */ }
```

Beautifying that output again yields the correct, stable result:

```js
var a = {
    /* c */
}
```

### Root cause

In `handle_end_block`, `empty_braces` was computed solely from `last_token` being the opening brace:

```js
var empty_braces = this._flags.last_token.type === TOKEN.START_BLOCK;
```

Comments held in `comments_before` are emitted without updating `last_token`, so a block whose only content is a comment was treated as a genuinely empty `{}` and no newline was printed before the closing brace. The symmetric `handle_start_block` already guards this case with `!next_token.comments_before`; this PR applies the same guard to `handle_end_block`:

```js
var empty_braces = !current_token.comments_before && this._flags.last_token.type === TOKEN.START_BLOCK;
```

Genuinely empty braces (`{}`) are unaffected and remain collapsed. The same change is applied to the Python implementation to keep the two implementations in parity.

### Tests

Two regression cases were added to the shared test data (`test/data/javascript/tests.js`), covering both an object literal and a block statement that contain only a block comment. `make ci` passes locally for both the JavaScript and Python implementations.

Fixes Issue: 

# Before Merge Checklist

- [ ] Added command-line option(s) (NA - no new option)
- [ ] README.md documents new feature/option(s) (NA - bug fix, no new feature/option)


sarathfrancis90:
Rebased onto current main; `make ci` passes locally (js + python suites, black clean).

Quick recap since it has been a while: a block containing only comments was treated as empty in `handle_end_block`, so the closing brace stayed on the comment line and a second pass moved it — beautify was not idempotent there. `handle_start_block` already guards the symmetric case with `!next_token.comments_before`; I applied the same check on the end side, so genuinely empty `{}` still collapses as before.

No rush — just flagging that it is current again if you want to take a look.

PR #2484, opened 2026-06-23
Fix bugs related to unicode identifiers python

GanerCodes:
# Description

Fixes Issue: #2411  

# Before Merge Checklist 


python/cssbeautifier/css/beautifier.py
```diff
@@ -1,5 +1,10 @@
 import sys
-import re
+
+# This is done due to the build system being how it is
+try:
+    re = __import__("regex")
+except ImportError:
+    import re
```
bitwiseman, line 7: Please explain further.

GanerCodes: [https://pypi.org/project/regex/](https://pypi.org/project/regex/) allows us to just use the character classes of ECMAScript identifiers directly. 
```py
_nonASCIIidentifierStartChars = r"\p{L}\p{Nl}"
_nonASCIIidentifierChars = r"\p{L}\p{Nl}\p{Nd}\p{Pc}\p{Mn}\p{Mc}"
```

PR #2506, opened 2026-07-05
fix: escape trailing whitespace before newlines in template literals …

Monu01123:
### Summary
Fixes **Issue #2390** where template literals containing real newlines and trailing horizontal whitespace get silently corrupted by editors that strip trailing whitespace on save (such as VS Code).

---

### Problem & Root Cause
When packers or code generators replace `\n` escape sequences with actual real newline characters inside template literals (backtick strings), `js-beautify` correctly preserves the real newlines. 

However, if the line ends with horizontal whitespace (spaces or tabs) immediately before that newline, `js-beautify` previously outputted that trailing whitespace as-is. When saved in an editor with "Strip Trailing Whitespace" enabled, those trailing characters are stripped away, silently altering the runtime string value of the template literal.

#### Example Input (packed code with actual newline instead of `\n`):
```javascript
let pattern=`[ \t
\f\r]`;

### CODE REVIEW

**Comment by bitwiseman** [comment]
Waiting on fix build failures.

File: test-fix-2390.js [hunk 1]
```diff
@@ -0,0 +1,70 @@
...
+'use strict';
+
+var beautify_js = require('./js/src/javascript/index');
+
+function test(description, input, expected) {
+  var result = beautify_js(input);
+  var pass = result === expected;
+  console.log((pass ? '✓' : '✗') + ' ' + description);
+  if (!pass) {
+    console.log('  INPUT   : ' + JSON.stringify(input));
```
bitwiseman: Use the built in `test/data/html/tests.js` file to test.

bitwiseman (changes requested):
@Monu01123 
Waiting for fixes.

